PR #417, opened 2023-08-12
Hides the stamp function.

momomomo111:
## Overview (Required)
- Stamp function can now be hidden.

## Links
- https://github.com/DroidKaigi/2023/issues/533

## Screenshot
Visible | Invisible
:--: | :--:
 |


core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched2023/data/remoteconfig/DefaultRemoteConfigApi.kt
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     * If you want to change the interval time to fetch, please change it here
+     */
+//    init {
+//        CoroutineScope(Dispatchers.IO).launch {
+//            firebaseRemoteConfig.settings {
+//                minimumFetchIntervalInSeconds = 12 * 3600L
+//            }
+//        }
+//    }
```
momomomo111, line 20: The original Firebase library has an implementation that automatically fetches changes when a value is updated, but the library for KMM does not, so the only way is to set `minimumFetchIntervalInSeconds` to fetch the value periodically.

PR #465, opened 2023-08-14
Remove constraintlayout-compose for Kotlin Multiplatform Compatibility

### LINKED ISSUE #270: Remove constraintlayout-compose for Kotlin Multiplatform Compatibility
Currently, constraintlayout-compose is not compatible with Kotlin Multiplatform, which is causing some issues in our project. One of the affected areas is BookmarkSheet. We need to remove constraintlayout-compose to maintain our project's compatibility with Kotlin Multiplatform.

fornewid:
## Issue
- close #270

## Overview (Required)
- Remove dependencies from all feature modules. 👋
- Use additional `Spacer` to align the layout without ConstraintLayout.

## Links
- None

## Screenshot
Before | After
:--: | :--:
 | 
 |


takahirom:
You are the first contributor after being public!!🏆

PR #479, opened 2023-08-14
Enable predictive back gesture for application scope

### LINKED ISSUE #474: [Suggestion] Add support for the predictive back gesture
**Idea Description**
The predictive back gesture is introduced in API level 33. Users will be able to preview whether or not they will back to the home screen when they do a back gesture and will be able to decide whether to stay in the current view or continue. Some applications (e.g., Settings, Spotify) are already introduced.
Under this project, there's no implementation requiring special consideration for `onBackPresed` logic handling. Thus, we can introduce the "predictive back" feature only for updating `AndroidManifest.xml`.

- https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture
- prototype implementation 
  - https://github.com/chigichan24/conference-app-2023/tree/enable-predictive-back
  - https://github.com/DroidKaigi/conference-app-2023/assets/7840108/1ccc287d-0d9c-460f-8799-e6f11b85f978

chigichan24:
## Issue
- close #474 

## Overview (Required)
- Enable predictive back gesture for application scope. We don't need to care about any back-pressed behavior due to there being no suspicious back-pressed logic handling.

## Links
- https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/app-android/src/main/AndroidManifest.xml
+++ b/app-android/src/main/AndroidManifest.xml
@@ -7,13 +7,14 @@
         android:name=".App"
         android:allowBackup="true"
         android:dataExtractionRules="@xml/data_extraction_rules"
+        android:enableOnBackInvokedCallback="true"
         android:fullBackupContent="@xml/backup_rules"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"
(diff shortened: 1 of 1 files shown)


takahirom (approved):
It's so easy to adapt! Thanks!

PR #482, opened 2023-08-14
[WIP]Address FIXME in Animation Calculation in BookmarkTopArea.kt

### LINKED ISSUE #273: Address FIXME in Animation Calculation in BookmarkTopArea.kt
The animation calculation in BookmarkTopArea.kt has a FIXME comment that needs to be addressed. Let's dive into the code, understand the problem, and provide a solution.

tfandkusu:
## Issue
- close #273

### Problem

- On this issue, `titlePaddingStart` and `titlePaddingTop` in the BookmarkTopArea.kt is calculated by `fraction`.
- `fraction` is culculated by `scrollState.firstVisibleItemScrollOffset` and fixed pixel 520.
- If user selects large font and bookmarks item which has long title, `scrollState.firstVisibleItemScrollOffset` may be more than 520 pixel.
- So `fraction` may be more than 1.0f.
- So `titlePaddingTop` may be less than 0.0f
- `Modifier.padding` must not be a negative value, or the app will crash.

### Solution

- I use first item's height insted of fixed pixel 520.
- First item's height is measured by `onGloballyPositioned` modifier

## Links

- [Jetpack Composeでサイズを取得する](https://zenn.dev/yumemi_inc/articles/36c43b84f8a3e3)

## Screenshot
Before(not use this `if` expresson) | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/16898831/d85063a7-ec13-4a2c-8c2b-4924519c7aac" width="300" /> | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/16898831/aa5326c5-27c0-4495-9e78-e5a4b3f735bf" width="300" />

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/BookmarkScreen.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/BookmarkScreen.kt
@@ -7,6 +7,9 @@ import androidx.compose.material3.Scaffold
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.testTag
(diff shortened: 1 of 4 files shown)


takahirom:
Thanks for delated investigation! Let me read the code and comments 👀

takahirom:
Could you tell me why we are using the height of the item? 🙋 I'm thinking about using the fixed size. 👀

takahirom:
I'm not sure if it's good to change the behavior based on the size of the first item. I feel that using a fixed height, we can provide a more consistent experience.

PR #493, opened 2023-08-14
fixed, BookmarkFilters chip color

### LINKED ISSUE #491: BookmarkFilters Chip has the wrong color applied to it.
In my previous work with https://github.com/DroidKaigi/conference-app-2023/pull/469, different colors were mapped to the BookmarkFilters Chip.

We need to check for the existence of Figma later to color it correctly.

According to the M3 guide, a Chip is a secondary container.

- https://m3.material.io/styles/color/the-color-system/color-roles

🙇🙇🙇

Pluu:
## Issue
- close #491

## Overview (Required)
- According to the M3 Color Rule, a chip should have a secondary container applied to it.
- Use secondary container Color
- The Light Theme colors you see in Figma are different from the actual Secondary colors, but that's a different issue, so we're only focusing on the color mapping.

## Links
- M3 : https://m3.material.io/styles/color/the-color-system/color-roles#904230ec-ae73-4f0f-8bff-4024a036ca66
- DroidKaigi figma : https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=54588-37352&mode=dev

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/BookmarkFilters.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/BookmarkFilters.kt
@@ -31,8 +31,8 @@ fun BookmarkFilters(
     modifier: Modifier = Modifier,
 ) {
     val selectedChipColor = AssistChipDefaults.assistChipColors(
-        containerColor = MaterialTheme.colorScheme.primary,
-        labelColor = MaterialTheme.colorScheme.onPrimary,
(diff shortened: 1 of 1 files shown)


tkhs0604 (approved):
Thank you for updating them! LGTM.

> The Light Theme colors you see in Figma are different from the actual Secondary colors, but that's a different issue, so we're only focusing on the color mapping.

Absolutely. It is related to the definition of design system. So let us make another issue about that 🙏

takahirom (approved):
great!

PR #508, opened 2023-08-15
Modify the design of the icons in the card section

### LINKED ISSUE #319: Modify the design of the icons in the card section on the timetable detail screen.
## Summary 
Modify the design of the icons in the card section

## Why
In the detail screen of the timetable item, there is a difference between the card component in the overview section and the design in Figma, which should be resolved.

| current situation | ideal situation |
| ---- | ---- |
| image |image |

## Goal
- [ ] Icon tint colors are different, change to appropriate color
- [ ] wording tint colors are different, change to appropriate color
- [ ] Change the background color of the card components to the appropriate color because they are different

## Reference
feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableItemDetailContent.kt

[Figma](https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=54902-39762&mode=design&t=8E9hM0o5yKWdKDoQ-4)

WonJoongLee:
## Issue
- close #319

## Overview (Required)
- Modified the design of the icons in the card section on the timetable detail screen.
- [x] Icon tint colors are different, change to appropriate color
- [x] wording tint colors are different, change to appropriate color
- [ ] Change the background color of the card components to the appropriate color because they are different

## Links
- [Figma link](https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=54902-39762&mode=design&t=8E9hM0o5yKWdKDoQ-4)

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableItemDetailSummaryCard.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableItemDetailSummaryCard.kt
@@ -12,11 +12,10 @@ import androidx.compose.material.icons.outlined.Place
 import androidx.compose.material.icons.outlined.Schedule
 import androidx.compose.material3.Card
 import androidx.compose.material3.CardDefaults
-import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Surface
-import androidx.compose.material3.surfaceColorAtElevation
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import io.github.droidkaigi.confsched2023.designsystem.theme.KaigiTheme
@@ -36,11 +35,7 @@ fun TimetableItemDetailSummaryCard(
     Card(
         shape = RoundedCornerShape(12.dp),
         modifier = modifier,
-        colors = CardDefaults.cardColors(
-            containerColor = MaterialTheme.colorScheme.surfaceColorAtElevation(
-                1.dp,
-            ),
-        ),
+        colors = CardDefaults.cardColors(containerColor = Color(0xFFF2F4F1))
     ) {
         Column(
             verticalArrangement = Arrangement.spacedBy(16.dp),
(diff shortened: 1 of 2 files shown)


feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableItemDetailSummaryCard.kt
```diff
@@ -36,11 +35,7 @@ fun TimetableItemDetailSummaryCard(
     Card(
         shape = RoundedCornerShape(12.dp),
         modifier = modifier,
-        colors = CardDefaults.cardColors(
-            containerColor = MaterialTheme.colorScheme.surfaceColorAtElevation(
-                1.dp,
-            ),
-        ),
+        colors = CardDefaults.cardColors(containerColor = Color(0xFFF2F4F1)),
```
takahirom, line 38: We might have to use theme color for applying a dark theme and it seems like we can apply it by adding elevation. 

image


> Surface at elevation +1 becomes Surface Container Low
https://material.io/blog/tone-based-surface-color-m3

The problem is how we can get elevation +1 😓 

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/ColorScheme.kt;l=478-484;drc=033f1c4e0e01fb2702b8f59520777aadaed46e5a

takahirom: You can leave this color as it is `surfaceColorAtElevation(1.dp,)` in this PR. In that case please create the issue for this 🙏

WonJoongLee: I think it would be better to leave the color as `surfaceColorAtElevation(1.dp,)`. I modified it at https://github.com/DroidKaigi/conference-app-2023/pull/508/commits/9db725f8719ea0e9c52d1bb3fe5e8adb692562dc . 

Also, I made the following issue for the card background color issue. ([link](https://github.com/DroidKaigi/conference-app-2023/issues/514))

takahirom (approved):
Thanks for your contribution and making issue!

PR #513, opened 2023-08-15
Applied transition to Stamps screen

### LINKED ISSUE #330: Apply transition to Stamps screen
Applied transition to Stamps screen

[Figma](https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=56147-22370&mode=dev)

momomomo111:
## Issue
- close #330

## Overview (Required)
- Transition to Stamps screen applied.
- This PR only implements screen transitions, and the base design will be addressed in a separate Issue.

## Screenshot
Before | After
:--: | :--:
 |


feature/stamps/src/main/java/com/momomomo111/stamps/StampsScreen.kt
```diff
@@ -0,0 +1,83 @@
...
+}
+
+fun NavController.navigateStampsScreen() {
+    navigate(stampsScreenRoute)
+}
+
+const val StampsScreenTestTag = "StampsScreen"
+
+@Composable
+fun StampsScreen(
```
takahirom, line 40: Could you add tests for StampsScreen for improving quality? 🙏 
Since contributors already know the feature so I think it is ok to add screenshot testing
You can refer this https://github.com/DroidKaigi/conference-app-2023/pull/409/files

momomomo111: Applied in this commit.

7fe77873e3c11b39916e2933e4b4ac84e469c0db

takahirom (approved):
Looks great! Thank you for creating the foundation of DroidKaigi!

feature/stamps/src/main/java/com/momomomo111/stamps/StampsScreen.kt
```diff
@@ -0,0 +1,83 @@
+package com.momomomo111.stamps
```
kwmt, line 1: Sorry for the sudden message 🙌 
I believe it would be better to change this package to `io.github.droidkaigi.confsched2023.stamps` like a [`about feature`](https://github.com/DroidKaigi/conference-app-2023/blob/main/feature/about/src/main/java/io/github/droidkaigi/confsched2023/about/AboutScreen.kt#L1).

takahirom: @momomomo111 Could you fix it? 😭

momomomo111: Sorry, the revised PR is here.🙏🙏🙏

https://github.com/DroidKaigi/conference-app-2023/pull/540

PR #523, opened 2023-08-15
[Proposal] add custom preview annotations for multi themes and languages

tkhs0604:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- We are adopting multi themes (light/dark) and languages (ja/en)
- So added annotations to preview these configurations

## Example

```xml
<!-- Japanese -->
<string name="aaa">あああ</string>

<!-- English -->
<string name="aaa">aaa</string>
```

```kotlin
@MultiThemePreviews
@MultiLanguagePreviews
@Composable
fun TestPreview() {
    KaigiTheme {
        Surface {
            Text(text = stringResource(id = R.string.aaa))
        }
    }
}
```

image

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/build-logic/src/main/kotlin/io/github/droidkaigi/confsched2023/primitive/KmpComposePlugin.kt
+++ b/build-logic/src/main/kotlin/io/github/droidkaigi/confsched2023/primitive/KmpComposePlugin.kt
@@ -34,6 +34,7 @@ class KmpComposePlugin : Plugin<Project> {
                         dependencies {
                             implementation(libs.findLibrary("androidxActivityActivityCompose").get())
                             implementation(libs.findLibrary("androidxLifecycleLifecycleRuntimeKtx").get())
+                            implementation(libs.findLibrary("composeUiToolingPreview").get())
                         }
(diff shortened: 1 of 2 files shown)


build-logic/src/main/kotlin/io/github/droidkaigi/confsched2023/primitive/KmpComposePlugin.kt
```diff
@@ -34,6 +34,7 @@ class KmpComposePlugin : Plugin<Project> {
                         dependencies {
                             implementation(libs.findLibrary("androidxActivityActivityCompose").get())
                             implementation(libs.findLibrary("androidxLifecycleLifecycleRuntimeKtx").get())
+                            implementation(libs.findLibrary("composeUiToolingPreview").get())
```
tkhs0604, line 37: Not sure if here is the right place to add a dependency 🤔

takahirom (approved):
Looks great! Thanks

PR #541, opened 2023-08-16
Apply stamps initial screen

### LINKED ISSUE #525: Apply stamps initial screen
**Description**
We would like you to implement the initial screen for the Stamps screen

image

[Figma](https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=55416-70904&mode=dev)

5altNaCl:
## Issue
- close #525 

## Overview (Required)
- Initial screen for stamps is implemented.
  - To implement, I've added the image of the stamp as a webp.

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
Sorry for bothering but we fixed the package name so please fix the conflict 🙇 
And can you debug why the test screenshot's first icon became big..? 🙏

5altNaCl:
@takahirom 
> Sorry for bothering but we fixed the package name so please fix the conflict 🙇 And can you debug why the test screenshot's first icon became big..? 🙏

OK, I also found some issues and will fix that too.
Turn the PR into Draft again until the fix is complete.

feature/stamps/src/main/java/io/github/droidkaigi/confsched2023/stamps/StampsScreen.kt
```diff
@@ -69,15 +83,98 @@ private fun StampsScreen(
...
+                Spacer(modifier = Modifier.height(24.dp))
+
+                Stamps(stamps = uiState.stamps)
             }
         },
     )
 }
+
+@Composable
+private fun Stamps(
```
takahirom, line 130: Please create a section as README says 🙏 
https://github.com/DroidKaigi/conference-app-2023#section

5altNaCl: OK, I'll try.

feature/stamps/src/main/java/io/github/droidkaigi/confsched2023/stamps/StampsScreen.kt
```diff
@@ -69,15 +85,123 @@ private fun StampsScreen(
...
-                Text(
-                    text = "Please implement StampsScreen!!!",
-                    style = MaterialTheme.typography.titleLarge,
-                )
+                item {
+                    StampsDetail()
+                }
+                item {
+                    Stamps(stamps = uiState.stamps, onStampsClick = onStampsClick)
+                }
```
momomomo111, line 104: Instead of passing an array with `item`, how about using `items` and passing each element of the array individually? Specifically, I think it would look something like the following (though the naming 'Stamp' might not be correct).
```suggestion
                items(uiState.stamps) { stamp ->
                    Stamp(stamp = stamp, onStampsClick = onStampsClick)
                }
```
If you find it difficult to configure the layout using `LazyColumn`, there's also the option of using [LazyVerticalGrid](https://developer.android.com/jetpack/compose/lists?hl=ja#lazy-grids).

5altNaCl: @momomomo111 
Sure, I think it would be simpler to display each stamp using `items` for the stamp.

However, I don't think there is enough information for the `stamps` variable to represent the state (valid or invalid) of each stamp from A to E.

Specifically, the display order of the stamps and the state of each stamp.

Therefore, in the new commit([45cb057](https://github.com/DroidKaigi/conference-app-2023/pull/541/commits/45cb0573a66f21357824a3ec307af2bad915712d)), we changed the status of stamps from A to E as Boolean as shown below, displaying valid if true and invalid if false.
image

Regarding the display order, I felt that it was immutable, so I used Column and Row to display it.

I think that isStampA~E needs to acquire the state from local storage in StampsScreenViewModel and reflect the result, or change it according to the NFC processing result that will be implemented in the future.

I still don't quite understand the logic of this app.
So, if you can point out what I'm doing wrong, or what I should fix, I'd appreciate it.

feature/stamps/src/main/java/io/github/droidkaigi/confsched2023/stamps/section/StampsSheet.kt
```diff
@@ -0,0 +1,38 @@
+package io.github.droidkaigi.confsched2023.stamps.section
+
+import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import io.github.droidkaigi.confsched2023.stamps.component.Stamps
+import io.github.droidkaigi.confsched2023.stamps.component.StampsDetail
+
+data class StampsSheetUiState(
+    val isStampA: Boolean,
```
takahirom, line 10: How about naming it `hasStampA`? Because UiState is not StampA 👀

feature/stamps/src/main/java/io/github/droidkaigi/confsched2023/stamps/StampsScreen.kt
```diff
@@ -69,15 +72,19 @@ private fun StampsScreen(
...
-                Text(
-                    text = "Please implement StampsScreen!!!",
-                    style = MaterialTheme.typography.titleLarge,
-                )
-            }
+            val layoutDirection = LocalLayoutDirection.current
+
+            StampsSheet(
+                modifier = Modifier.padding(
+                    PaddingValues(
```
takahirom, line 79: Maybe we don't have to use PaddingValues and you can use "Modifier.padding(top =xxx, start=xxx" but we can make a issue for this so it's up to you if we do it or not. 👍

feature/stamps/src/main/res/drawable-hdpi/img_stamp_a_off.webp
takahirom, line 1: Webp 👍

takahirom (commented):
Almost LGTM! Could you check naming? 🙏

feature/stamps/src/main/java/io/github/droidkaigi/confsched2023/stamps/StampsScreen.kt
takahirom: 👀

jmatsu: Android Lint prints an error when ignoring the padding from Scaffold by the way.

5altNaCl: @takahirom 
Thank you for your comment.
As I read it, I felt that PaddingValues is indeed redundant.
So I thought it would be better to do something like this:

```
modifier = Modifier.padding(
    top = padding.calculateTopPadding() + 20.dp,
    start = padding.calculateStartPadding(layoutDirection) + 16.dp,
    end = padding.calculateEndPadding(layoutDirection) + 16.dp,
)
```

However, the point to note here is that the bottom padding is not specified.
Specifying bottom padding causes unnecessary space between BottomNavigation and Screen.

feature/stamps/src/main/java/io/github/droidkaigi/confsched2023/stamps/section/StampsSheet.kt
5altNaCl: @takahirom 
Certainly, the naming seems to be good!

feature/stamps/src/main/java/io/github/droidkaigi/confsched2023/stamps/StampsScreen.kt
jmatsu: Oh, I misunderstood the first comment's intention. Yes, instantiating PaddingValues is truly redundant. Sorry for confusing you all!

takahirom: So maybe all you have to do is remove PaddingValues and leave it as it is for other codes 👀

feature/stamps/src/main/java/io/github/droidkaigi/confsched2023/stamps/StampsScreen.kt
momomomo111: @5altNaCl 
Thank you for your suggestion!
If we set `isStampA` to `true`, we would need to re-compose not only the Compose function related to isStampA but also other places, which would not be efficient in terms of performance.
How about creating a `data class` and using `items` to reduce the coupling between stamps?
The `data class` of Sponsor might serve as a reference.

```kotlin
package io.github.droidkaigi.confsched2023.model

data class Stamp(
    val visibleImageResName: String,
    val hiddenImageResName: String,
    val isStampVisible: Boolean = false,
    val contentDescription: String
)
```
`example use for Android`
```kotlin
val stamp = Stamp("img_stamp_hoge_on", "img_stamp_hoge_off", true, "hoge")
val visibleImageResId = context.resources.getName(stamp.visibleImageResName)
```

`example use for iOS`
```swift
let stamp = KMMModule.Stamp(visibleImageResName: "img_stamp_hoge_on", hiddenImageResName: "img_stamp_hoge_off", isStampVisible: true, accessibilityDescription: "hoge")
let visibleImage = UIImage(named: stamp.visibleImageResName)
```

Thank you for bringing up the number of stamps!
Considering the potential increase in the number of stamps in the future, we should take that into account.
I apologize for the lack of clarity in the specifications. How about using [LazyVerticalGrid](https://developer.android.com/jetpack/compose/lists?hl=ja#lazy-grids) to display a maximum of two stamps side by side?
I assume you might be concerned about the display order, but it will be arranged according to the array's index, so it should be fine🙆‍♂️.

takahirom: I think we don't use Compose Multiplatform for stamp so we don't think we have to use res name 👀

momomomo111: @5altNaCl 
The initial reason for proposing ResName was to accommodate KMM. However, it seems that this is not necessary, so there is no need to use ResName. I apologize for the confusion.🙏

takahirom (approved):
Looks great!

momomomo111 (approved):
Thank you very much! Looking good! 💫

PR #555, opened 2023-08-16
Display Snackbar for Bookmark Confirmation in TimetableItemDetailScreen #517

### LINKED ISSUE #517: Display Snackbar for Bookmark Confirmation in TimetableItemDetailScreen
**Idea Description**
When a user adds a bookmark in the TimetableItemDetailScreen, a Snackbar should appear with the message "Bookmarked successfully."

This enhancement will provide immediate feedback to the user, confirming that the action was successful.

image

KoutaMatsushita:
## Issue
- close #517 

## Overview (Required)
- Show snackbar after the bookmark added.

## Links
-

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableItemDetailViewModel.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableItemDetailViewModel.kt
@@ -9,7 +9,6 @@ import io.github.droidkaigi.confsched2023.model.SessionsRepository
 import io.github.droidkaigi.confsched2023.model.TimetableItem
 import io.github.droidkaigi.confsched2023.model.TimetableItemId
 import io.github.droidkaigi.confsched2023.sessions.strings.TimetableItemDetailStrings
-import io.github.droidkaigi.confsched2023.ui.UserMessageResult.Dismissed
 import io.github.droidkaigi.confsched2023.ui.UserMessageStateHolder
 import io.github.droidkaigi.confsched2023.ui.buildUiState
 import io.github.droidkaigi.confsched2023.ui.handleErrorAndRetry
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableItemDetailViewModel.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableItemDetailViewModel.kt
@@ -68,9 +68,6 @@ class TimetableItemDetailViewModel @Inject constructor(
                     TimetableItemDetailStrings.BookmarkedSuccessfully.asString(),
                     TimetableItemDetailStrings.ViewBookmarkList.asString(),
                 )
-            } else {
-                val currentMessage = userMessageStateHolder.messageUiState.userMessages.firstOrNull() ?: return@launch
-                userMessageStateHolder.messageShown(currentMessage.id, Dismissed)
             }
         }
     }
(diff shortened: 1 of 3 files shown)


feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableItemDetailViewModel.kt
```diff
@@ -60,6 +62,16 @@ class TimetableItemDetailViewModel @Inject constructor(
...
             sessionsRepository.toggleBookmark(timetableItem.id)
+            val bookmarked = timetableItemStateFlow.value?.second ?: return@launch
+            if (bookmarked) {
+                userMessageStateHolder.showMessage(
+                    TimetableItemDetailStrings.BookmarkedSuccessfully.asString(),
+                    TimetableItemDetailStrings.ViewBookmarkList.asString(),
+                )
+            } else {
+                val currentMessage = userMessageStateHolder.messageUiState.userMessages.firstOrNull() ?: return@launch
+                userMessageStateHolder.messageShown(currentMessage.id, Dismissed)
```
KoutaMatsushita, line 73: When the button is tapped, the Snackbar disappears, but at that time, the Dissmiss event does not occur, so I manually dismiss it 😢

takahirom: "The button" means "View" button or Bookmark button?

KoutaMatsushita: Bookmark button.
When the bookmark button is tapped, the Snackbar disappears, but it seems that `userMessageStateHolder.messageShown` is not being called. 
Do you have any idea why this might be happening?

KoutaMatsushita: I tried, but it was the same userMessage.

```kotlin
    userMessageStateHolder.messageUiState.userMessages.firstOrNull()?.let { userMessage ->
        println(userMessageStateHolder.messageUiState.userMessages) // ← added
        LaunchedEffect(userMessage) {
```

first: 
```
[UserMessage(message=ブックマークに追加されました, actionLabel=一覧を見る, id=-5097040755527889007, userMessageResult=null)]
```

second: 
```
[UserMessage(message=ブックマークに追加されました, actionLabel=一覧を見る, id=-5097040755527889007, userMessageResult=null), UserMessage(message=ブックマークに追加されました, actionLabel=一覧を見る, id=2112774890980393949, userMessageResult=null)]
```

In this pattern, it seems that `snackbarHostState.showSnackbar` is not returning `SnackbarResult`. As a result, the suspend function in the `filter` gets stuck, and `userMessages` is not removed.

https://github.com/DroidKaigi/conference-app-2023/blob/86faf75d8b35cc190e912b7230cf105ec47ab497/core/ui/src/androidMain/kotlin/io/github/droidkaigi/confsched2023/ui/UserMessage.kt#L73-L89

 This causes `firstOrNull` to yield the same result.

takahirom: Can you dive into snackbarHostState.showSnackbar? Or can you provide a reproducing branch? 🙏

KoutaMatsushita: OK. I will give it a try, so please wait for a while.（or use this branch https://github.com/KoutaMatsushita/conference-app-2023/tree/bookmark_confirmation_snackbar ）

KoutaMatsushita: SnackbarHost is disappearing due to recompose. 😓 
suppress recompose using `remember { SnackbarHostState() }`.

takahirom: I'll keep that in mind 😭 Thanks for investigating!

takahirom:
Thanks for investigating and fixing it!

PR #571, opened 2023-08-16
Remove "interpretation target" chip and add another lang chip when the session is interpretation target

### LINKED ISSUE #463: Remove "interpretation target" chip and align the chip to the Figma
**Describe the bug**

Timetable screen doesn't working well with BigFont/BigDisplay config.

- The bottom of header content hidden by the sheet.
- Chips can be second line if needed.
- 

**To Reproduce**
1. Go to Timetable screen

**Expected behavior**

The hidden UI should be visible, otherwise, can be consider with designer as well.

**Screenshots**

tfandkusu:
## Issue
- close #463

## Overview (Required)

In the `TimetableListItem` of `TimetableList`

- When session is interpretation target, labels are `EN` `JA`
- When session is English only, label is `EN`
- When session is Japanese only, label is `JA`
- When session is Mixed, label is `MIXED`. There is no Mixed session, but I adapted the existing implementation.

## Links
- 

## Screenshot
Before | After
:--: | :--:
 | 
 |


tfandkusu:
I updated Overview, Screenshot and implementation.

mhidaka:
Brilliant! Thank you for your quick commits! We'll take a p-r review time tomorrow. 💪💪💪

core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2023/model/TimetableLanguage.kt
```diff
@@ -4,10 +4,11 @@ public data class TimetableLanguage(
     val langOfSpeaker: String,
     val isInterpretationTarget: Boolean,
 ) {
-    val langLabel = when (langOfSpeaker) {
-        "MIXED" -> langOfSpeaker
-        else -> {
-            langOfSpeaker.take(2)
-        }
+    val labels = if (langOfSpeaker == Lang.MIXED.tagName) {
```
takahirom, line 7: Looks great

takahirom (approved):
Thanks for your cool code! 🆒

PR #552, opened 2023-08-16
Add placeholder to search text field

hiraike32:
## Issue
- close #550 

## Overview (Required)
- Add placeholder to search text field
- Create compose preview for easy checking

## Links
- [Figma](https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=55052%3A45676&mode=dev)

## Screenshot
Before | After | After(English)
:--: | :--: | :--:
 |  |

PR #574, opened 2023-08-16
Implement feature that displays results according to search input text

### LINKED ISSUE #524: Search Feature Not Working: Empty Screen Displayed Regardless of Input
The current search feature in our application appears to be malfunctioning. Despite entering text into the search field, the result is always an empty screen.

### Description
When users attempt to search by entering text, the expected search results are not displayed. Instead, an empty screen appears every time, making the search feature unusable.

chigichan24:
## Issue
- close #524

## Overview (Required)
In this PR, I focused on the below implementations.
- Update Empty view UI (show search query on empty view)
- Show search text field hint.
- Show search result timetable items which include it can move to the session detail screen and can do bookmark action.
- Apply edge-to-edge for the search result page.

### The difference of TimetableListItem
 On Figma, there're no UI guides for search results.  Therefore, basically, I follow to TimetableList implementation. But I updated 1. Add hour information for each session item 2. Add a day label. 

difference of TimetableListItem

I can follow to designer's or maintainer's suggestion. Please feedback on my update.🙇‍♂️

### Out of this PR scope
- Additional filter conditions (for now, it works "days" filter only)
- (Refactoring) Extract common components used in `BookMarkList`, `TimetableList`, and `SearchList`.
- Detail spec for search UI (due to there being no UI guides for this screen)

## Links
- https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=54462-36316&mode=design

## Screenshot
Before | After
:--: | :--:
 | 
 | 
 |


chigichan24:
Oops. It seems the PR change will conflict with #552 . I'll rebase it after #552 is merged.

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/SearchScreenViewModel.kt
```diff
@@ -78,4 +118,10 @@ class SearchScreenViewModel @Inject constructor(
             },
         )
     }
+
+    fun updateBookmark(timetableItem: TimetableItem) {
```
takahirom, line 122: It's my preference but how about using `onClickTimetableItemBookmark` to promote the best practice that notifying events to ViewModel.

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableListItem.kt
```diff
@@ -58,7 +61,7 @@ fun TimetableListItem(
             .clickable { onClick(timetableItem) },
     ) {
         Row(verticalAlignment = Alignment.CenterVertically) {
-            Row(modifier = Modifier.weight(1F)) {
+            FlowRow(modifier = Modifier.weight(1F)) {
```
takahirom, line 64: 👍

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/section/SearchList.kt
```diff
@@ -0,0 +1,144 @@
...
+import kotlinx.collections.immutable.PersistentList
+import kotlinx.collections.immutable.PersistentSet
+import java.util.Locale
+
+@Composable
+fun SearchList(
+    contentPaddingValues: PaddingValues,
+    scrollState: LazyListState,
+    bookmarkedTimetableItemIds: PersistentSet<TimetableItemId>,
+    timetableItems: PersistentList<TimetableItem>,
```
takahirom, line 49: It is a very small point. As I mentioned here we would like to have UiState for each section. 🙏 
https://github.com/DroidKaigi/conference-app-2023#section

takahirom (approved):
Looks great! Please fix the conflict!

takahirom (approved):
Your code looks cool! 🆒 
Thanks

PR #579, opened 2023-08-16
Add Filtering Feature by Category, Session Type, and Supported Language.

### LINKED ISSUE #384: Implement filtering feature except "event date"
## What is required
The filtering feature is being implemented but only for "event date" now. So let's implement for the other conditions.
258847872-63d0efb4-f19e-447d-81d5-167230a5c1a0

## Reference
https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=54918-45437&mode=design&t=FstLHmTOBbCbVZhz-0

kitakkun:
## Issue
- close #384 

### View Changes
- Implement Chip component for Session Type, and Supported Language filters (`FilterSessionTypeChip`, `FilterLanguageChip`).
- Replace `Row` in `SearchFilter` composable function with `LazyRow` to make it scrollable. Also changed the way to add padding to avoid view clipping issue.

### Internal Changes
- Add `TimetableSessionType` enum class to localize session types' strings.
- Add `sessionType` member variable to `TimetableItem` class.
- Add `SessionType` string resource to `SessionStrings`. (JA: `"セッション種別"`, EN: `"Session Type"`)
- Observe flow of `Timetable` in `SessionRepository` at `SearchScreenViewModel` to obtain available `categories` and `sessionTypes`.

## Links
- https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=54918-45437&mode=design&t=FstLHmTOBbCbVZhz-0

### English UI demo
Before | After
:--: | :--:
 | 

### Japanese UI demo
Before | After
:--: | :--:
 |


chigichan24:
@kitakkun 
JFYI: This PR may conflict with my change #574 . Keep in mind~🙇‍♂️

kitakkun:
I've just noticed `Filter` data class has a member `filterFavorites` variable(may be same as bookmark). Should I add `BookmarkChip` to filter by Bookmark state? (Though it doesn't exist in the figma design.)

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/SearchScreenViewModel.kt
```diff
@@ -122,6 +115,32 @@ class SearchScreenViewModel @Inject constructor(
...
+                } else {
+                    remove(sessionType)
+                }
+            },
+        )
+    }
+
+    fun onLanguagesSelected(language: Lang, isSelected: Boolean) {
+        val selectedLanguages = searchFilterUiState.value.selectedLanguages.toMutableList()
+        searchFilterUiState.value = searchFilterUiState.value.copy(
```
takahirom, line 133: I don't think we should copy the UiState because this broke the single source of UiState 💭 
But this is existence implementation so I'm going to create issue and it is ok to leave it as it is 👍 
https://github.com/DroidKaigi/conference-app-2023#single-source-of-truth-with-builduistate-

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/FilterLanguageChip.kt
```diff
@@ -0,0 +1,93 @@
...
+    selectedLanguages: List<Lang>,
+    languages: List<Lang>,
+    onLanguagesSelected: (Lang, Boolean) -> Unit,
+    modifier: Modifier = Modifier,
+    isSelected: Boolean = false,
+    selectedLanguagesValues: String = "",
+    onFilterLanguageChipClicked: () -> Unit,
+) {
+    var expanded by remember { mutableStateOf(false) }
+    val onLanguagesSelectedUpdated by rememberUpdatedState(newValue = onLanguagesSelected)
```
takahirom, line 38: I'm not sure if we should use rememberUpdatedState here. Is there any reason why you used rememberUpdatedState while other click events don't use this 👀

takahirom:
Please fix this pr's conflict 🙏

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/FilterLanguageChip.kt
kitakkun: Just because FilterDayChip was implemented like that. I think just invoke the passed lambda function `onLanguageSelected` is okay... 🤔

takahirom (approved):
There may be room for improvement but it is certainly a good step to start!

PR #580, opened 2023-08-16
Changed to not include Firebase implementation on the iOS side

momomomo111:
## Overview (Required)
- Changed to not include Firebase implementation on the iOS side.

## Links
- [Slack](https://droidkaigi.slack.com/archives/C05236NLCNL/p1692192920165159)


core/data/src/androidMain/kotlin/io/github/droidkaigi/confsched2023/data/di/ApiModule.kt
```diff
@@ -159,6 +160,8 @@ class RemoteConfigModule {
     @Provides
     @Singleton
     fun provideFirebaseRemoteConfig(): RemoteConfigApi {
-        return DefaultRemoteConfigApi()
```
takahirom, line 162: We can move DefaultRemoteConfigApi to androidMain and you can leave this as it is

momomomo111: I was able to confirm the operation and get a normal value.

07e9400877a2c4afc63436ba3f73a0c299db6a3e

takahirom: It's not for Andorid but for iOS 🙏 
When iOS use DefaultRemoteConfigApi, DefaultRemoteConfigApi don't have the implementation class right?
Can you check this?
./gradlew :app-ios-shared:assembleSharedReleaseXCFramework

core/data/src/iosMain/kotlin/io/github/droidkaigi/confsched2023/data/remoteconfig/DefaultRemoteConfigApi.kt
```diff
@@ -0,0 +1,12 @@
+package io.github.droidkaigi.confsched2023.data.remoteconfig
+
+import io.github.droidkaigi.confsched2023.data.remoteconfig.FakeRemoteConfigApi.Status.Default
+
+actual class DefaultRemoteConfigApi : RemoteConfigApi {
```
takahirom, line 5: I don't think we should make this class because iOS engineers will use this and won't notice if they use fake👀

takahirom (approved):
Anyway it is better to work. Let me merge this

PR #593, opened 2023-08-17
Display SideEvent List on FloorMapScreen

### LINKED ISSUE #421: Display SideEvent List on FloorMapScreen
**Idea Description**
Add SideEvent items to the FloorMapScreen to complete the list. Place the list's Composable functions in the `section` package, and create the UiState accordingly. More details can be found [here](https://github.com/DroidKaigi/conference-app-2023#section).

image
https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=55876-19767&mode=dev

morayl:
## Issue
- close #421 

## Overview (Required)
- Implemented FloorMapSideEventItem
- Changed to null one of fake(SideEvents)'s link for confirming
- Commenting the code because I have a question about colors and icons

## Screenshot
Before | After(with or without link)
:--: | :--:
none |


feature/floor-map/src/main/java/io/github/droidkaigi/confsched2023/floormap/component/FloorMapSideEventItem.kt
```diff
@@ -1,18 +1,146 @@
...
+        }
+        Spacer(modifier = Modifier.height(16.dp))
+        Divider()
+    }
+}
+
+@Composable
+private fun Mark.iconResAndColor(): Pair<Int, Color> {
+    return when (this) {
+        Mark1 -> R.drawable.ic_favorite to LocalContentColor.current
```
morayl, line 112: I tried to define the color of the icon, but couldn't find the right place.
For example, I thought it would be good to define pink in designsystem/theme/Color.kt? (It didn't seem to be visible in the visibility of the module)
In the sense of temporary, if you directly define #DC369A here, it will be resolved.

feature/floor-map/src/main/res/drawable/ic_favorite.xml
morayl, line 1: I found out that it is better to use "Icons", so I will fix it.

feature/floor-map/src/main/java/io/github/droidkaigi/confsched2023/floormap/component/FloorMapSideEventItem.kt
```diff
@@ -1,18 +1,149 @@
...
+        }
+        Spacer(modifier = Modifier.height(16.dp))
+        Divider()
+    }
+}
+
+@Composable
+private fun Mark.iconResAndColor(): Pair<ImageVector, Color> {
+    return when (this) {
+        Mark1 -> Icons.Filled.Favorite to LocalContentColor.current
```
morayl, line 115: I tried to define the color of the icon, but couldn't find the right place.
For example, I thought it would be good to define pink in designsystem/theme/Color.kt? (It didn't seem to be visible in the visibility of the module)
In the sense of temporary, if you directly define #DC369A here, it will be resolved.

core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2023/model/SideEvent.kt
```diff
@@ -53,7 +53,7 @@ val SideEvents = persistentListOf(
             enTitle = "DAY1-DAY2 10:00-11:00",
         ),
         mark = Mark1,
-        link = "https://github.com/DroidKaigi/conference-app-2023",
+        link = null,
```
takahirom, line 56: 👍

feature/floor-map/src/main/java/io/github/droidkaigi/confsched2023/floormap/component/FloorMapSideEventItem.kt
```diff
@@ -1,18 +1,149 @@
...
+                contentDescription = FavoriteIcon.asString(),
+                modifier = Modifier.size(16.dp)
+            )
+            Spacer(modifier = Modifier.width(8.dp))
+            Text(
+                text = sideEvent.timeText.currentLangTitle,
+                style = MaterialTheme.typography.bodySmall,
+            )
+        }
+        if (sideEvent.link != null) {
```
takahirom, line 89: 👍

feature/floor-map/src/main/java/io/github/droidkaigi/confsched2023/floormap/component/FloorMapSideEventItem.kt
```diff
@@ -1,18 +1,149 @@
...
+            )
+        ) {
+            append(FloorMapStrings.EventDetail.asString())
+        }
+    }
+}
+
+@Preview
+@Composable
+fun PreviewFloorMapSideEventItem() {
```
takahirom, line 140: 💯

feature/floor-map/src/main/java/io/github/droidkaigi/confsched2023/floormap/component/FloorMapSideEventItem.kt
takahirom: I would like to have it in Mark properties. This is because it is better that the member of adding SideEvent doesn't have to know this when expression

```kotlin
enum MarkColors {
  Pink,
  ...
}
```

morayl: @takahirom 
Thanks！
But I don't fully understand.

like this?

```
SideEvent {...
    enum class Mark(val color: MarkColor) {
        Mark1(Pink),
    }

    enum class MarkColor {
        Pink;
    }
```

```
// In FloorMapSideEventItem.kt

            Icon(
                imageVector = iconVector,
                contentDescription = FavoriteIcon.asString(),
                tint = sideEvent.mark.color,
                modifier = Modifier.size(16.dp),
            )
```
sideEvent.mark.color is just enum.
We have to associate the enum with the color, should I associate it with the FloorMapSideEventItem?

```
// In FloorMapSideEventItem.kt(In some function)
MarkColor.Pink -> Color(0xFFF8FAF6)
```

I think it would be better to unify the pink color to one value, but is it okay to define the value (0xFFF8FAF6) directly in the "Floor map module"?

takahirom: All I want to do is that DroidKaigi staff can add an side event easily.
I though they should modify here when they do it.
I'll give it a thought.👀

takahirom: I was wondering if you could create an issue for binding Marks and Color and icons. 🙇 
I realized this is what we have to do.

takahirom (approved):
Thanks for your contribution!

feature/floor-map/src/main/java/io/github/droidkaigi/confsched2023/floormap/component/FloorMapSideEventItem.kt
morayl: I've got it and I agree.
But sorry I just push 0b02a52 and a342450.

And I think it' better bind ColorObject and ColorHex which doesn't define at Color.kt companion object to avoid different modules may have different pink colors.(I just come up with an idea)

takahirom (approved):
Looks great! Thanks for your contribution and cooperation!

PR #595, opened 2023-08-17
[feature/#545] Resolve Stacking Issue in Bottom Navigation When Pressing Timetable Items

### LINKED ISSUE #545: Fix Stacking Issue in Bottom Navigation When Pressing Timetable Items
### Problem Description
Pressing the timetable items in the bottom navigation leads to the same screen stacking on top of each other. This makes it impossible for the user to navigate back, causing a frustrating experience.

### Suggested Solution
We need to identify the cause of this issue and implement a fix to ensure that pressing a timetable item doesn't create redundant screens. This may involve adjusting the navigation logic within the app.

### Additional Context
This issue is critical for maintaining a smooth navigation experience within the app. Your insights and contributions to solving this problem are highly appreciated!

image

l2hyunwoo:
## Issue
- close #545 

## Overview (Required)
- Add launchSingleTop option when navigate to other screen in bottom navigation

### DIFF
--- a/app-android/src/main/java/io/github/droidkaigi/confsched2023/KaigiApp.kt
+++ b/app-android/src/main/java/io/github/droidkaigi/confsched2023/KaigiApp.kt
@@ -205,7 +205,10 @@ class KaigiAppMainNestedGraphStateHolder : MainNestedGraphStateHolder {
             Timetable -> mainNestedNavController.navigateTimetableScreen()
             About -> mainNestedNavController.navigateAboutScreen()
             FloorMap -> mainNestedNavController.navigateFloorMapScreen()
-            Contributor -> mainNestedNavController.navigate(contributorsScreenRoute)
+            Contributor -> mainNestedNavController.navigate(contributorsScreenRoute) {
+                launchSingleTop = true
+                restoreState = true
+            }
             Badges -> mainNestedNavController.navigateStampsScreen()
         }
     }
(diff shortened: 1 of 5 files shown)


app-android/src/main/java/io/github/droidkaigi/confsched2023/KaigiApp.kt
```diff
@@ -205,7 +205,10 @@ class KaigiAppMainNestedGraphStateHolder : MainNestedGraphStateHolder {
             Timetable -> mainNestedNavController.navigateTimetableScreen()
             About -> mainNestedNavController.navigateAboutScreen()
             FloorMap -> mainNestedNavController.navigateFloorMapScreen()
-            Contributor -> mainNestedNavController.navigate(contributorsScreenRoute)
+            Contributor -> mainNestedNavController.navigate(contributorsScreenRoute) {
+                launchSingleTop = true
+                restoreState = true
+            }
```
l2hyunwoo, line 211: It's different from other cases. Should it fix?

takahirom: This screen is created to investigate the Compose Multiplatform. so we are not using navigation for now. I think we can have it in androidMain though.

takahirom (approved):
Looks great! Thanks for creating infrastructure of DroidKaigi app

PR #599, opened 2023-08-17
Modify about screen's title

hiraike32:
## Issue
- none

## Overview (Required)
- Fixed the title of the About screen to match Figma's design when scrolling it's screen.

## Links
- [Figma](https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=56147%3A41182&mode=dev)

## Screenshot
Before | After | Figma
:--: | :--: | :--:
before | after | スクリーンショット 2023-08-17 14 08 54

### DIFF
--- a/feature/about/src/main/java/io/github/droidkaigi/confsched2023/about/AboutScreen.kt
+++ b/feature/about/src/main/java/io/github/droidkaigi/confsched2023/about/AboutScreen.kt
@@ -8,16 +8,24 @@ import androidx.compose.foundation.layout.calculateEndPadding
 import androidx.compose.foundation.layout.calculateStartPadding
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Scaffold
 import androidx.compose.material3.SnackbarHost
 import androidx.compose.material3.SnackbarHostState
+import androidx.compose.material3.Text
+import androidx.compose.material3.TopAppBar
+import androidx.compose.material3.TopAppBarDefaults
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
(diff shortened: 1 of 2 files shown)


feature/about/src/main/java/io/github/droidkaigi/confsched2023/about/AboutScreen.kt
```diff
@@ -81,8 +90,29 @@ private fun AboutScreen(
     versionName: String?,
     onLinkClick: (url: String) -> Unit,
 ) {
+    val scrollBehavior = TopAppBarDefaults.pinnedScrollBehavior()
     Scaffold(
         modifier = Modifier.testTag(AboutScreenTestTag),
+        topBar = {
+            TopAppBar(
+                title = {
+                    if (scrollBehavior.state.overlappedFraction == 0f) {
```
hiraike32, line 99: Makes the title larger when the scroll is at the top.

feature/about/src/main/java/io/github/droidkaigi/confsched2023/about/AboutScreen.kt
```diff
@@ -81,8 +90,29 @@ private fun AboutScreen(
...
+                        Text(
+                            text = AboutStrings.Title.asString(),
+                            style = MaterialTheme.typography.headlineLarge,
+                            fontWeight = FontWeight.Medium,
+                        )
+                    } else {
+                        Text(
+                            text = AboutStrings.Title.asString(),
+                            color = Color.Unspecified.copy(alpha = scrollBehavior.state.overlappedFraction),
```
hiraike32, line 108: When scrolled, the title is made smaller and gradually displayed according to the scroll position.
I wanted to use LargeTopAppBar if possible, but I couldn't create a good layout on the About screen because the icon part at the top was in the way.
Therefore, TopAppBar is used to simulate the behavior of LargeTopAppBar by controlling the transparency.

takahirom (approved):
Looks great! Thanks for great user experience!

PR #617, opened 2023-08-17
Hide session title at session detail screen when enable the BigFont/BigDisplay

### LINKED ISSUE #464: Hide session title at session detail screen when enable the BigFont/BigDisplay
**Describe the bug**

The session title hide the top of text when enable the BigFont/BigDisplay

For some of screen are affected, then investigate the cause first?

**To Reproduce**
1. Go to Session detail screen

**Expected behavior**

Showing the session title correctly.

**Screenshots**

kosenda:
## Issue
- close #464

## Overview (Required)
- The title of the session detail screen is not broken when using BigFont/BigDisplay.

## Problems that could not be resolved
When the title is one line, the size is smaller than the size that does not cause character breakage.
Before | After
:--: | :--:
 | 

#### [Android developers - Paragraph](https://developer.android.com/reference/kotlin/androidx/compose/ui/text/Paragraph)
- Reference for use

#### [AutoSizing TextField in Jetpack compose](https://blog.canopas.com/autosizing-textfield-in-jetpack-compose-7a80f0270853)
- Refer to how to use Paragraph

#### [Paragraph]
I was going to use `MultiParagraph` at first, but [here](https://developer.android.com/reference/kotlin/androidx/compose/ui/text/MultiParagraph#:~:text=class%20MultiParagraph-,Lays,-out%20and%20renders)
I used `Paragraph` because I did not need multiple styles in the text as shown in

#### [LargeTopAppBar height]
I couldn't figure out what to compare the height of the Paragraph with to calculate whether or not a character break occurs, even after looking at the documentation and reference sites.
So, I checked the actual values as I entered them.
First, I looked at Figma and targeted the height of the headline (72.dp), which did not cause broken text, but reduced the text size to the limit.
Next, I set the height of LargeTopAppBar (152.dp), which is described in Material Design, and found that the text was cut off.
Next, I set the height of LargeTopAppBar (152.dp) minus the padding (28.dp) at the bottom of LargeTopAppBar (124.dp), and the text was not broken.
Next, setting it to 128.dp resulted in broken characters, so we used 124.dp above for comparison.

### **FontSize and DisplaySise is Max**

Phone | Before | After
:--: | :--: | :--:
Pixel6 (h: 892dp, w: 412.dp) | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/60963155/40cd5ec8-2ea7-44ef-8621-c6e74ce4289e" width="300" /> | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/60963155/75d68b39-0014-4cd1-9bf5-7fa8eb21eb55" width="300" />
Emulator Small Phone(h: 640dp, w: 360.dp) | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/60963155/6e1e3436-d5c8-4e4d-9b0c-9b659c13a11b" width="300" /> | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/60963155/54a4363e-ede4-459e-b5ff-fbfb6c844a81" width="300" />


feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableItemDetailScreenTopAppBar.kt
```diff
@@ -82,20 +92,48 @@ private fun ResizeableText(
...
+                constraints = constraints,
+                maxLines = maxLines,
+                ellipsis = true,
+            )
+        }
+
+        // calculate before displaying to avoid broken character.
+        var intrinsics = calculateParagraph()
+        while (intrinsics.height > maxHeight.value / density.density && fontSize > minFontSize.value) {
+            println("ログ fontSize: $fontSize, styleIndex: $styleIndex, intrinsics.height: ${intrinsics.height}")
```
takahirom, line 119: You may forget removing the log 😄

kosenda: I'm Sorry!!

takahirom:
I'm not sure if this work.. 💭 
Can you provide `PreviewTimetableItemDetailScreenTopAppBar` to see it?
image

kosenda:
I understand 🫡

kosenda:
I have created a fontScale and a long or short preview, does it meet your expectations?

takahirom:
Sorry for bothering you but it looks like our test system doesn't support the preview parameter.
Can you provide a separated method?
I'm going to create an issue for it 🙇

takahirom:
I had a misunderstanding. But PreviewParameter seems work but FontScalePreviews doesn't work. So please implement separated methods 🙇

kosenda:
The font scale doesn't seem to work if I just separate...

takahirom:
Thanks for adding tests. I was writing something like this. 

```kotlin
class PreviewTimetableItemDetailScreenTopAppBarTitleAndFontScaleProvider :
    PreviewParameterProvider<Pair<MultiLangText, Float>> {
    override val values: Sequence<Pair<MultiLangText, Float>> = sequenceOf(
        Session.fake().title,
        Session.fake().title.copy(jaTitle = Session.fake().title.jaTitle.repeat(5)),
    ).flatMap { title -> listOf(0.5F, 1.0F, 1.5F, 2.0F).map { title to it } }
}

@OptIn(ExperimentalMaterial3Api::class)
@Composable
internal fun PreviewTimetableItemDetailScreenTopAppBar(
    @PreviewParameter(
        PreviewTimetableItemDetailScreenTopAppBarTitleAndFontScaleProvider::class,
    ) titleAndFontScale: Pair<MultiLangText, Float>,
) {
    val (title, scale) = titleAndFontScale
    val newConfiguration = LocalConfiguration.current.apply {
        this.fontScale = scale
    }
    CompositionLocalProvider(LocalConfiguration provides newConfiguration) {
        KaigiTheme {
            Surface {
                TimetableItemDetailScreenTopAppBar(
                    title = title,
                    onNavigationIconClick = {},
                    scrollBehavior = TopAppBarDefaults.exitUntilCollapsedScrollBehavior(),
                )
            }
        }
    }
}
```

kosenda:
I deleted test and set the code you raised above.

takahirom:
It seems it doesn't work 😭

kosenda:
ですね…🥲
テストに戻します

takahirom:
I thought the text should be small. What do you think?
image

kosenda:
Using an emulator, we confirmed the broken text on a device (Nexus One) similar to the snap shot test.
So we will try to make it smaller!

kosenda:
## Memo
Perhaps BoxWithConstraints is not working ...

I`m thinking of changing my size comparison from 
```
intrinsics.height > maxHeight.value / density.density
```
to 
```
intrinsics.height > LARGE_TOP_APP_BAR_HEIGHT_DP
```
I checked on NexusOne with LARGE_TOP_APP_BAR_HEIGHT_DP set to 124 and no broken characters occurred.

kosenda:
I've done my best, but if the quality can't be merged, I'd like to give up.

takahirom:
OK! How about just adding the test?

kosenda:
Was I correct in only adding the test and deleting the other code?

takahirom:
But I mean so

kosenda:
Sorry for everything 🙇.
I will leave only the test and delete the other codes!

kosenda:
I am thinking of closing this pull request and creating the test-only pull request again, is that ok?

takahirom:
OK! Thanks for everything!

PR #614, opened 2023-08-17
Revert "Update AGP"

atsushieno:
This reverts commit 02d5f5f5f18ca79750a8653772233a8f1a899488.

This fixes #612.

## Issue
- close #612.

## Overview (Required)

See #612.

PR #619, opened 2023-08-17
Add per-app language support.

### LINKED ISSUE #618: Let me use the app in Japanese on my English system
**Idea Description**

Android 13 makes it possible for us to choose the language we want to use for each app. The feature is however opt-in and we have to make minor changes in the app. My Android device defaults to English, but since the majority of those sessions are in Japanese and the session titles are most likely determined in Japanese first, I would like to run it in Japanese mode.

I have a patch that makes it.

https://developer.android.com/guide/topics/resources/app-languages

atsushieno:
Fixes #618

A few thing to complement:

- I explicitly specify `ja` and `en-US` as they are indeed the supported languages.
- I did not resort to [automatic per-app language support](https://developer.android.com/guide/topics/resources/app-languages#auto-localeconfig) as it depends on AGP 8.1 which I just reverted in #614 :p

## Issue
- close #618

## Overview (Required)

See #618.

## Screenshot

### DIFF
--- a/app-android/src/main/AndroidManifest.xml
+++ b/app-android/src/main/AndroidManifest.xml
@@ -14,6 +14,7 @@
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
         android:theme="@style/Theme.KaigiApp"
+        android:localeConfig="@xml/locales_config">
         tools:targetApi="tiramisu">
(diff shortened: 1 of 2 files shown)


takahirom (approved):
Looks great! Thanks!

PR #637, opened 2023-08-17
Use the entries property instead of the values function of enum classes

tatsuyafujisaki:
## Issue
- none

## Overview (Required)
- Use the `entries` property instead of the `values()` function of enum classes for performance.

## Links
- https://kotlinlang.org/docs/whatsnew19.html#stable-data-objects-for-symmetry-with-data-classes


takahirom (approved):
It is better and sound! Thanks!

PR #647, opened 2023-08-17
Simplify ReadMoreOutlinedButton Click State Change

### LINKED ISSUE #626: Simplify the change of state when ReadMoreOutlinedButton is clicked
**Idea Description**
When we click the ReadMoreOutlinedButton, we check its current state to change the value. However, it might be simpler to just set it directly to `false`

https://github.com/DroidKaigi/conference-app-2023/blob/c27309eab4f5a252f8c282e6e667eabe2b9c4ee2/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableItemDetailContent.kt#L95

sanao1006:
## Issue
- close #626

## Overview (Required)
- When ReadMoreOutlinedButton is pressed, change isExpanded directly to true


takahirom (approved):
Thanks!

PR #651, opened 2023-08-17
Update implement onSideEventClick event

### LINKED ISSUE #645: Implement navigation to url of side event
**Idea Description**
Implement onSideEventClick event. Please open with chrome custom tab following existing implementation.

スクリーンショット 2023-08-17 23 11 08

umsys:
## Issue
- close #645 

## Overview (Required)
I implemented the onSideEventClick event.
I modified the variable passed by the onSideEventClick event to be of String type with only a URL. Upon clicking, it now opens in Chrome's custom tab, following the existing implementation.

## Links
- 

### Before 
https://github.com/DroidKaigi/conference-app-2023/assets/45583423/6bb489c3-7db7-4bcc-8c41-9a029ea49515
### After
https://github.com/DroidKaigi/conference-app-2023/assets/45583423/c8b92e8e-0f89-4677-a01c-22f298e1e0a8

### DIFF
--- a/app-android/src/main/java/io/github/droidkaigi/confsched2023/KaigiApp.kt
+++ b/app-android/src/main/java/io/github/droidkaigi/confsched2023/KaigiApp.kt
@@ -165,9 +165,7 @@ private fun NavGraphBuilder.mainScreen(
                 onLinkClick = externalNavController::navigate,
             )
             nestedFloorMapScreen(
-                onSideEventClick = {
-                    TODO()
-                },
+                onSideEventClick = externalNavController::navigate,
             )
             nestedStampsScreen(
                 onStampsClick = {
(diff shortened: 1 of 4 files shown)


umsys:
[ask]
Should I have implemented the onSideEventClick only on the link text "Event_detail" part in the following code, instead of the entire item?
https://github.com/umsys/conference-app-2023/blob/ed1135e7e13b959192bf428b85808df9afff87f6/feature/floor-map/src/main/java/io/github/droidkaigi/confsched2023/floormap/component/FloorMapSideEventItem.kt#L91-L108

umsys:
@upon0426
Would you be able to make a decision regarding this matter?

takahirom:
I think it is OK. There are some points to consider.
1. Minimizing dependency
2. Type safe
3. Customizable by the SideEvent
But I don't think using string is a bad idea.

takahirom (approved):
Thanks for considering what we should do!

PR #657, opened 2023-08-17
Add screenshot test for changing floors of the floor map

### LINKED ISSUE #624: Add Screenshot Test for Changing Floors on Floor Map
**Idea Description**
Implement a screenshot test to verify the functionality of changing floors on the Floor Map.

https://github.com/DroidKaigi/conference-app-2023/blob/main/feature/floor-map/src/test/java/io/github/droidkaigi/confsched2023/FloorMapScreenTest.kt#L35

image

maxfie1d:
## Issue
- close #624 

## Overview (Required)
- Added screenshot test for changing floors of the floor map to 1F

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/floor-map/src/test/java/io/github/droidkaigi/confsched2023/FloorMapScreenTest.kt
+++ b/feature/floor-map/src/test/java/io/github/droidkaigi/confsched2023/FloorMapScreenTest.kt
@@ -36,6 +36,8 @@ class FloorMapScreenTest {
         floorMapScreenRobot {
             setupScreenContent()
             checkScreenCapture()
+            setFloorLevelToGround()
+            checkScreenCapture()
         }
     }
 }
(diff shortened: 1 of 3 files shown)


feature/floor-map/src/test/java/io/github/droidkaigi/confsched2023/FloorMapScreenTest.kt
```diff
@@ -36,6 +36,8 @@ class FloorMapScreenTest {
         floorMapScreenRobot {
             setupScreenContent()
             checkScreenCapture()
+            setFloorLevelToGround()
+            checkScreenCapture()
```
takahirom, line 40: Thanks. Almost LGTM! But I would like to add what we want to check to the file name. So could you add test method and maybe in that test you don't have to take two screenshot.

maxfie1d: I extracted the test method https://github.com/DroidKaigi/conference-app-2023/pull/657/commits/88cfd0dd81201e66ff8dfe9ecd74314eec5e9958. Please let me know if it's not what you intended 👋🏼

takahirom (approved):
Looks great! Thanks for your coorperation!

PR #661, opened 2023-08-18
Show image if FloorMapSideEventItem's image link exists

### LINKED ISSUE #646: FloorMapSideEventItem should accept image if exists
**Idea Description**
Update FloorMapSideEventItem so that it can to show image if exists.

スクリーンショット 2023-08-17 23 13 53

matsumo0922:
## Issue
- closes #646

## Overview (Required)
- Added `ImageUrl` field to SideEvent data class
- Added dummy data for ImageUrl
- Added Image composable to be displayed when there is an ImageUrl

## Question
- Is `SideEvent` currently only dummy data?
- If there is something like DesignDoc, I will follow it, but...

## Screenshot
Before | After | figma
:--: | :--: | - 
 |  |

### DIFF
--- a/feature/floor-map/src/main/java/io/github/droidkaigi/confsched2023/floormap/component/FloorMapSideEventItem.kt
+++ b/feature/floor-map/src/main/java/io/github/droidkaigi/confsched2023/floormap/component/FloorMapSideEventItem.kt
@@ -16,6 +16,7 @@ import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Favorite
 import androidx.compose.material.icons.filled.Link
+import androidx.compose.material.icons.filled.Person
 import androidx.compose.material.icons.filled.Schedule
 import androidx.compose.material3.Divider
 import androidx.compose.material3.Icon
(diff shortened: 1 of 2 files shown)


feature/floor-map/src/main/java/io/github/droidkaigi/confsched2023/floormap/component/FloorMapSideEventItem.kt
```diff
@@ -52,55 +60,77 @@ fun FloorMapSideEventItem(
...
+                Image(
+                    modifier = Modifier
+                        .size(80.dp)
+                        .clip(RoundedCornerShape(16.dp))
+                        .border(
+                            width = 1.dp,
+                            color = MaterialTheme.colorScheme.outline,
+                            shape = RoundedCornerShape(16.dp),
+                        ),
+                    painter = rememberAsyncImagePainter(sideEvent.imageLink!!),
```
takahirom, line 131: Could you use `previewOverride`? Because this causes flaky screenshot tests since it involves a network request.

matsumo0922: sorry, missed it. c26a7f5

takahirom (approved):
Looks great! Thanks for your contribution!

PR #677, opened 2023-08-18
:recycle: Make session text selectable.

### LINKED ISSUE #607: Make session text selectable
**Idea Description**
When we find something interesting word we want to search for it. So having a feature we can copy would be convenient!
image

Corvus400:
## Issue
- close #607 

## Overview (Required)
- Session detail text can now be selected using SelectionContainer.
- I would have liked to implement a Context Menu other than Copy, but at this point it seems to be impossible with the standard features.

## Links
- https://developer.android.com/jetpack/compose/text?hl=ja
- https://issuetracker.google.com/issues/218266808

## Movie

https://github.com/DroidKaigi/conference-app-2023/assets/13657682/e4381c85-a36e-41a5-97ef-efd56f1dfacd

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableItemDetailContent.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableItemDetailContent.kt
@@ -16,6 +16,7 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.foundation.text.selection.SelectionContainer
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Person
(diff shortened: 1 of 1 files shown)


feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableItemDetailContent.kt
```diff
@@ -82,22 +83,24 @@ private fun DescriptionSection(
...
-            color = MaterialTheme.colorScheme.onSurface,
-            maxLines = if (isExpanded) Int.MAX_VALUE else 5,
-            overflow = TextOverflow.Ellipsis,
-            modifier = Modifier.padding(start = 16.dp, end = 16.dp),
-        )
-        if (!isExpanded) {
-            ReadMoreOutlinedButton(
-                onClick = { isExpanded = true },
-                modifier = Modifier.padding(top = 16.dp, start = 16.dp, end = 16.dp),
+    SelectionContainer {
```
takahirom, line 86: I didn't know this component! Thanks

PR #689, opened 2023-08-18
Remove a redundant `suspend` modifier

woxtu:
## Issue
- close N/A

## Overview (Required)
- This PR removes a redundant `suspend` modifier and resolves the following build warning:

```
w: file:///***/conference-app-2023/core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched2023/data/NetworkService.kt:16:16 Redundant 'suspend' modifier: lambda parameters of suspend function type uses existing continuation.
```


takahirom:
Thanks for fixing the warning!

PR #690, opened 2023-08-18
Add blur effect around switcher

### LINKED ISSUE #643: Add blur effect around switcher
**Idea Description**
Implement blur container with gradation around switcher

スクリーンショット 2023-08-17 22 49 30

taratara10:
## Issue
- close #643 

## Overview (Required)
- Implement blur container with gradation around switcher.

261329511-342a6ce3-e08e-4595-82dd-44f1222d3364

### Light Theme
Before | After
:--: | :--:
 | 

### Dark Theme
Before | After
:--: | :--:
 |


feature/floor-map/src/main/java/io/github/droidkaigi/confsched2023/floormap/FloorMapScreen.kt
```diff
@@ -138,6 +142,18 @@ private fun FloorMapScreen(
                             .padding(bottom = 56.dp),
                     )
                 }
+                Spacer(
+                    modifier = Modifier
+                        .fillMaxWidth()
+                        .height(96.dp)
+                        .align(Alignment.BottomCenter)
+                        .background(
+                            brush = Brush.verticalGradient(
```
takahirom, line 151: There is a `Brush.verticalGradient` already here. Could we use it?
This might be not an easy problem 👀 
https://github.com/DroidKaigi/conference-app-2023/pull/690/files#diff-631fb12247c515279975a2569c2ac6255f80753353c8b7a97b071696c2b8a6c1R137

taratara10: Thanks for pointing that out!
The `fadingEdge` was not visible, so I fixed it.

In the previous order of modifier, blur was drawn first and then paddig was applied.The blur was not visible because it was hidden by the padding.
So, I set the padding first and then draw the bluer in the display area, the code is now as intended!


Before | After(Light Theme)|After(Dark Theme)|
:--: | :--:| :--:
 | |

takahirom:
Looks better thanks!

PR #693, opened 2023-08-18
[iOS] add required dependency to timetable module

tkhs0604:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- Build for iOS fails in my environment (not sure if it happens only my local though 🤔).
- The build issue has been resolved by adding the dependency to Component from Timetable.

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |


ry-itto (approved):
Good catch! Thanks!

PR #694, opened 2023-08-18
Use modifier timetable screens

### LINKED ISSUE #556: TimetableScreen's modifier parameter is not used
**Idea Description**
We have to pass the modifier parameter to the root component of the composable function

kako351:
## Issue
- close #556

## Overview (Required)
- TimetableScreen modifier parameter is used.

## Screenshot

- There's no change in appearance

### DIFF
--- a/app-android/src/main/java/io/github/droidkaigi/confsched2023/KaigiApp.kt
+++ b/app-android/src/main/java/io/github/droidkaigi/confsched2023/KaigiApp.kt
@@ -143,9 +143,9 @@ private fun NavGraphBuilder.mainScreen(
 ) {
     mainScreen(
         mainNestedGraphStateHolder = KaigiAppMainNestedGraphStateHolder(),
-        mainNestedGraph = { mainNestedNavController, padding ->
+        mainNestedGraph = { mainNestedNavController, _ ->
             nestedSessionScreens(
-                modifier = Modifier.padding(padding),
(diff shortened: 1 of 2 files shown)


app-android/src/main/java/io/github/droidkaigi/confsched2023/KaigiApp.kt
```diff
@@ -143,9 +143,9 @@ private fun NavGraphBuilder.mainScreen(
 ) {
     mainScreen(
         mainNestedGraphStateHolder = KaigiAppMainNestedGraphStateHolder(),
-        mainNestedGraph = { mainNestedNavController, padding ->
+        mainNestedGraph = { mainNestedNavController, _ ->
             nestedSessionScreens(
-                modifier = Modifier.padding(padding),
+                modifier = Modifier,
```
kako351, line 148: Using the modifier in TimetableScreen created unwanted space at the bottom, as shown in the screenshot below.
This was caused by padding in this area, and when the padding was removed, the unwanted space disappeared.

If this padding is something that should not be removed, please comment. 🙏

takahirom: I think it should be removed🙏

RyuNen344 (approved):
L G t M ~ !! Thanks for your contribution!!

PR #697, opened 2023-08-18
chore: introduce complete kotlin for windows user

RyuNen344:
## Issue
- close nothing

## Overview (Required)
- complete kotlin gradle plugin allows windows users to see Darwin APIs
- ref: https://github.com/LouisCAD/CompleteKotlin

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |


RyuNen344:
@takahirom 
What do you think about support for android developer that uses windows PC?

takahirom:
I've stared this repository.
But I didn't know how it works. But I read the README. It looks great!

>In a nutshell, it downloads files that contain platform klibs intended for other OSes, and puts them in the right directories. It only downloads the missing ones, so it will have no effect on macOS if you don't use Windows (aka. mingw) or Linux MIPS targets for example.

takahirom:
> What do you think about support for android developer that uses windows PC?

I'm not sure we can support it appropriately because I don't have it. But we can try to support at least.

takahirom (approved):

takahirom:
@RyuNen344 
I've approved. Please merge this when you want.

PR #699, opened 2023-08-18
[Build] Use Gradle's build cache and/or configuration cache properly

### LINKED ISSUE #56: [Build] Use Gradle's build cache and/or configuration cache properly
The current codebase don't use Gradle's cache system properly so some tasks like Lint are fully executed. The build time will be significantly decreased if we can properly use Gradle builds.

masah517:
## Issue
- close #56

### - Configuration cache
Regarding of the configuration cache, I have enable it by adding properties into "gradle.properties".
In terms of verification, I have run the following gradle task twice, and see the configuration cache was reused with the proper log.

Command 
`./gradlew assembleDebug`

**Result**

1st time run
configuration_cache_first_run

2nd time run
configuration_cache_second_run

configuration cache availability
configuration_cache_error_check

### - Build cache
Regarding of the Gradle's build cache, it seems to have been implemented as the following. So please correct me if I am missing something, or any further implementation is required here(e.g. Change the default cache folder path)

https://github.com/DroidKaigi/conference-app-2023/blob/7a9f0544126251b789733621624de56d8aaf48c1/gradle.properties#L4

**Result**

Command to check Gradle's build cache location
` ls -a  $HOME/.gradle/caches`

Command to clear Gradle's build cache
`rm -rf  $HOME/.gradle/caches/build-cache-*   `

Screenshot
build_cache

## Links
Reference links
- [configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:usage:enable)
- [build cache](https://docs.gradle.org/current/userguide/build_cache.html#sec:build_cache_enable)

## Screenshot
Before | After
:--: | :--:
before_configuration_cache| after_configuration_cache


gradle.properties
```diff
@@ -3,6 +3,9 @@ org.gradle.jvmargs=-Xmx2048m -Dfile.encoding=UTF-8
 org.gradle.configureondemand=true
 org.gradle.caching=true
 org.gradle.parallel=true
+# configuration-cache
+org.gradle.unsafe.configuration-cache=true
+org.gradle.unsafe.configuration-cache-problems=warn
```
RyuNen344, line 8: @masah517 
Thanks for trying!!

`configuration-cache-problems=warn` causes a problem that the error does not occur when the configuration cache is ineffective.
Could you try enable `configuration cache` without `problems=warn` option ?

FYI: AGP and MPP gradle plugin has already support configuration cache 😉
(maybe we need to migration convention plugin to support configuration cache)

masah517: @RyuNen344 
Will try as advised! Thank you

masah517: @RyuNen344 
Hi there!
It seems like the build keeps failing even though I remove `problems=warn` option.

May I please have support on what you mean by 
> maybe we need to migration convention plugin to support configuration cache)

Thank you

jmatsu: Yes, of course. Enabling configuration caches properly needs 4 steps.

- Turn the problem level to warn.
- Support confiugration caches for owned custom plugins including convention plugins by following https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:testing.
- Turn it back to error.
- Until the build passes, specify Task#notCompatibleWithConfigurationCache for 3rd party tasks that haven't supported caches.

Please note that it's not the beat approach that just turns configuration cache problems into warn level. The option helps developers to support caches gradually but not a fisrt choice for enabling caches by plugin users. If you have to do it, ` org.gradle.configuration-cache.max-problems=<small mumber>` must to be set to control your projects.

RyuNen344:
sorry for late, we will check ❤️

masah517:
@RyuNen344 
Noted! Thank you!

jmatsu (changes requested):
Awesome! Could you please answer to my question and rebase to the latest main branch? Thank you.

build-logic/src/main/kotlin/io/github/droidkaigi/confsched2023/primitive/KmpAndroidPlugin.kt
```diff
@@ -34,6 +36,9 @@ class KmpAndroidPlugin : Plugin<Project> {
                     }
                 }
             }
+            tasks.withType<KspTaskMetadata>().configureEach {
+                notCompatibleWithConfigurationCache("Configuration cache not supported due to serialization")
```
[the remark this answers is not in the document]
masah517: @jmatsu 
Sorry for the late reply. 
Regrading of the issue, I couldn't find any official announcement about it. 

However I have commented the link of Ksp slack channel that someone ran into the same issue.
[Link](https://slack-chats.kotlinlang.org/t/13166064/been-discovering-that-the-task-kspcommonmainkotlinmetadata-i#9a50fa1b-1ec5-47c2-9172-2a5780a1900e   )

Plz lmk if you any concerns. Thank you.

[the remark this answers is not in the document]
jmatsu: Sure, that's enough! 👍

jmatsu:
Now waiting for the CI jobs...

jmatsu (approved):
Great work! Thank you! 🎉

PR #706, opened 2023-08-18
:recycle: ClickableLinkText was made common and used on each screen.

Corvus400:
## Issue
- none

## Overview (Required)
- Commonized ClickableLinkText.
- The commonized parts were used in the following screens
1. Session Details Screen
2. Floor Map Screen
3. About Screen

## Movie

after

PR #708, opened 2023-08-18
(proposal) :sparkles: Implement Translate Button.

Corvus400:
# Attention
- Please review this PR after #706 is merged as it uses the content of #706.

## Issue
- none

## Overview (Required)
- I thought it would be useful to have a translation function on the session detail screen.
- This is accomplished by encoding the Description into Google Translate and attaching it to the text query.
- We decided to use this method because we thought it would not increase the size of the application and would not require the implementation of a new API.

## Movie

https://github.com/DroidKaigi/conference-app-2023/assets/13657682/cd822dea-15f8-4b36-b525-528708884382

### DIFF
--- a/app-android/src/main/java/io/github/droidkaigi/confsched2023/KaigiApp.kt
+++ b/app-android/src/main/java/io/github/droidkaigi/confsched2023/KaigiApp.kt
@@ -106,6 +106,7 @@ private fun KaigiNavHost(
                     timetableItem.id,
                 )
             },
+            onLinkClick = externalNavController::navigate,
         )
         searchScreen(
             onNavigationIconClick = {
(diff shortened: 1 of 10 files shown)


takahirom:
Thanks for your proposal. DroidKaigi has our global team and some staff involving this. We are discussing this issue.

takahirom:
Status: We already have the translation so maybe we can provide translation by ourselves. (Just we don't have time to implement it to the server) But we might use this design.

takahirom:
I think now we have i18nDesc in the response. I'll talk with staff

takahirom:
@Corvus400 
The designer says that "A" would be best and "B" is also good.
image

Corvus400:
@takahirom 
Yes, sir!
Is it safe to start implementation with proposal A?

takahirom:
@Corvus400 Yes! Go ahead!

takahirom:
Let me close this since there is a new pr 🙏

PR #725, opened 2023-08-19
Quit copying SearchScreenViewModel's filter UiState to encourage SSoT of UiState

### LINKED ISSUE #670: Quit copying SearchScreenViewModel's filter UiState to encourage SSoT of UiState
**Idea Description**
Please refer to the link.
It may be like this

Current

```kotlin
private val filterUiState = MutableStateFlow()

val uiState = buildUiState(x,filterUiState) {
  UiState(
  filterUiState.copy()
}


fun onXxx() {
  filterUiState.value = filterUiState.value.copy(xxx)
}
```

What I want to have

```kotlin
private val filterStateFlow = MutableStateFlow()

val uiState = buildUiState(x,filter) { x, filterStateFlow ->
  UiState(
  FilterUiState(
    filter.xxxx
  )
}


fun onXxx() {
  filterStateFlow.value = filterStateFlow.value.copy(xxx)
}
```

https://github.com/DroidKaigi/conference-app-2023/pull/579/files#r1297306479
https://github.com/DroidKaigi/conference-app-2023#single-source-of-truth-with-builduistate-

Hachimori:
## Issue
- close #670

## Overview (Required)
- In `SearchScreenViewModel`, it now privately hosts filter's state (instead of search filter's Ui state) for filtering sessions.
  - The change is aimed for following Single Source of Truth principle.

## Links
- 

## Screenshot
- Before

https://github.com/DroidKaigi/conference-app-2023/assets/3529042/2c22da47-43f5-4504-a66a-0fecce796454

- After (There shouldn't be any difference)

https://github.com/DroidKaigi/conference-app-2023/assets/3529042/36bb996a-c786-48de-8541-ba949b3eb057

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/SearchScreenViewModel.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/SearchScreenViewModel.kt
@@ -47,35 +47,44 @@ class SearchScreenViewModel @Inject constructor(
             initialValue = Timetable(),
         )
 
-    private val searchFilterUiState: MutableStateFlow<SearchFilterUiState> = MutableStateFlow(
-        SearchFilterUiState(),
+    private val filterStateFlow: MutableStateFlow<Filters> = MutableStateFlow(
+        Filters(),
     )
 
     val uiState: StateFlow<SearchScreenUiState> = buildUiState(
         searchQuery,
-        searchFilterUiState,
+        filterStateFlow,
         sessionsStateFlow,
-    ) { searchQuery, searchFilterUiState, sessions ->
+    ) { searchQuery, filters, sessions ->
         val searchedSessions = sessions.filtered(
             Filters(
-                days = searchFilterUiState.selectedDays,
-                categories = searchFilterUiState.selectedCategories,
-                sessionTypes = searchFilterUiState.selectedSessionTypes,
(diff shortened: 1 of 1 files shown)


feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/SearchScreenViewModel.kt
```diff
@@ -47,35 +47,44 @@ class SearchScreenViewModel @Inject constructor(
...
-                days = searchFilterUiState.selectedDays,
-                categories = searchFilterUiState.selectedCategories,
-                sessionTypes = searchFilterUiState.selectedSessionTypes,
-                languages = searchFilterUiState.selectedLanguages,
                 searchWord = searchQuery,
+                days = filters.days,
+                categories = filters.categories,
+                sessionTypes = filters.sessionTypes,
+                languages = filters.languages,
             ),
```
Hachimori, line 66: It might be better to implement like `filters.copy(searchWord = searchQuery)` for sake of simplicity, but I wondered that it might violate Single Source of Truth principle 🤔

takahirom: I don't want to copy UiState for creating consistent UI but filter is not UiState so I think it is ok.

takahirom (approved):
There is a small point but it is better! Thanks!

PR #727, opened 2023-08-19
Rename StampsList

### LINKED ISSUE #726: Change the name from StampsList to StampList
**Idea Description**
The name 'StampsList' implies a list that returns multiple data sets, so I suggest changing it to 'StampList'. What do you think?

https://github.com/DroidKaigi/conference-app-2023/blob/fae756b1935c29d0293cd0e8e7b2ece9ea6f0770/feature/stamps/src/main/java/io/github/droidkaigi/confsched2023/stamps/section/StampsList.kt#L21-L24

momomomo111:
## Issue
- close #726

## Overview (Required)
- Given that the name 'StampsList' seems to imply a list returning multiple data sets, I've changed it to 'StampList'.


takahirom:
🙏

Issue #460, opened 2023-08-14
Improve foldable devices support for TimeTable screen.


fumiya-kume:
**Idea Description**

The current experiences with the foldable devices is enough, However we can consider to optimize it as well. In fact. Google want to push the app that optimize it, then We can try to the edge as first penguin!

Ref: https://android-developers.googleblog.com/2023/07/introducing-new-play-store-for-large-screens.html

As offer a basis for discussion, I'd like to recommend to apply the `List-detail` pattern for the screen. since the Time table screen has Main and detail screen then.

I know that require discussion and making the design, but I believe it will good benefit for the Android developers.

Ref: https://m3.material.io/foundations/layout/canonical-layouts/list-detail

**Reference images and links**
 
Current experiences at Timetable screen *Pixel Fold
<img src="https://github.com/DroidKaigi/conference-app-2023/assets/16269075/c2b706a5-6103-43f9-bfa4-a65324676869" width="50%" />

takahirom:
Thank you for the suggestion, @fumiya-kume!

The idea of implementing the List-detail pattern for the TimeTable screen is indeed interesting, and it seems to align well with our goal to enhance the experience for foldable devices. I would like to discuss this approach with the designer to ensure that it integrates well with the overall vision for the app.

Your input is incredibly valuable, and I appreciate you taking the initiative to propose this change. Let's continue to collaborate as we explore this idea further.

fumiya-kume:
Thank you for picking up my feedback 🙇 

In addition, `Navigation rail` is working well with foldable devices.
Definitely, better than `NavigationBar` / `BottomNabigationBar` when using the foldable devices.

Ref: https://m3.material.io/components/navigation-rail/overview

l2hyunwoo:
If assignee is empty, I want to try this issue.

upon0426:
Sure! Thank you for showing the interest! Assigned you!

l2hyunwoo:
@upon0426 I think it can consider to separate two parts.

- Phase 1: Apply Navigation Rail when foldable device expanded
- Phase 2: Apply List-Detail Pattern

Make two tasks in one PR can cause decreased readability and increased complexity. So I think make two PRs in each phase is better to proceed. Would it be okay to do that?

upon0426:
Thank you for the consideration!
Yes, separate PR sounds good to me as well for this kind of case.
Please proceed that way 👍

takahirom:
[...]

PR #738, opened 2023-08-19
[feature/#460] Set NavigationRail when display expanded or device is tablet

l2hyunwoo:
## Overview (Required)
- Phase 1: Apply Navigation Rail when foldable device expanded(or device is tablet)
- I think UI test should be added to next PR, what do you think about it? 
  - If UI test is essential when ui is modified, I'll include UI test code in this.

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/main/src/main/java/io/github/droidkaigi/confsched2023/main/MainScreen.kt
+++ b/feature/main/src/main/java/io/github/droidkaigi/confsched2023/main/MainScreen.kt
@@ -1,6 +1,7 @@
 package io.github.droidkaigi.confsched2023.main
 
 import androidx.annotation.DrawableRes
+import androidx.compose.animation.AnimatedVisibility
 import androidx.compose.animation.EnterTransition
 import androidx.compose.animation.ExitTransition
 import androidx.compose.animation.core.FastOutLinearInEasing
@@ -10,7 +11,9 @@ import androidx.compose.animation.fadeIn
 import androidx.compose.animation.fadeOut
 import androidx.compose.animation.scaleIn
 import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.WindowInsets
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.CalendarMonth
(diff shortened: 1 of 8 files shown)


takahirom:
Looks great. Could you add test something like this?

@Config(qualifiers = RobolectricDeviceQualifiers.MediumTablet)
...
@Test
fun checkMediumTabletLaunchShot()...

l2hyunwoo:
> Looks great. Could you add test something like this?
> ```
> @Config(qualifiers = RobolectricDeviceQualifiers.MediumTablet)
> ...
> @Test
> fun checkMediumTabletLaunchShot()...
> ```

It's in new commit!

feature/main/src/main/java/io/github/droidkaigi/confsched2023/main/MainScreen.kt
```diff
@@ -63,22 +78,36 @@ interface MainNestedGraphStateHolder {
...
 }
 
+enum class NavigationType {
+    BOTTOM_NAVIGATION, NAVIGATION_RAIL
+}
+
 @Composable
 fun MainScreen(
+    windowSize: WindowSizeClass,
+    displayFeatures: ImmutableList<DisplayFeature>,
```
takahirom, line 88: What do you think you are going to use this? 👀

feature/main/src/main/java/io/github/droidkaigi/confsched2023/main/MainScreen.kt
```diff
@@ -137,34 +166,49 @@ data class MainScreenUiState(
...
-            startDestination = "timetable",
-            modifier = Modifier.padding(padding),
-            enterTransition = { materialFadeThroughIn() },
-            exitTransition = { materialFadeThroughOut() },
-        ) {
-            mainNestedNavGraph(mainNestedNavController, padding)
+        }
+        Scaffold(
+            bottomBar = {
+                AnimatedVisibility(visible = navigationType == BOTTOM_NAVIGATION) {
```
takahirom, line 190: 👍

feature/main/src/main/java/io/github/droidkaigi/confsched2023/main/MainScreen.kt
l2hyunwoo: 1) windowSize: to calculate WindowWidthSizeClass and apply proper navigationType(in this pr) and content pane type(in next PR, apply dual-pane(as known as List-Detail Pattern) or single-pane)

2) displayFeatures: When WindowWidthSizeClass is Medium, content arrangement type should be determined by device folding posture. (according to material 3 sample application [reply](https://github.com/android/compose-samples/blob/main/Reply/app/src/main/java/com/example/reply/ui/ReplyApp.kt#L99))

l2hyunwoo: As [explained](https://github.com/DroidKaigi/conference-app-2023/issues/460#issuecomment-1684697989), list-detail pattern will apply in this screen. So `displayFeatures` will be used to implement it.

takahirom: Great! Thanks

takahirom (approved):
Looks awesome!

PR #750, opened 2023-08-19
[iOS] Try add xcodebuild for ios build workflow

ry-itto:
This reverts commit 2c264a68ab252a91056736b6de1e019c77776e16.

## Issue
- close #ISSUE_NUMBER

- This PR might be working... https://github.com/DroidKaigi/conference-app-2023/pull/693

## Overview (Required)
- 

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |


ry-itto:
@tomoya0x00 
This seems like working. Please check this? 🙏🏼

app-ios/Modules/Package.swift
```diff
@@ -106,6 +106,7 @@ var package = Package(
             name: "Sponsor",
             dependencies: [
                 "Assets",
+                "Component",
```
tomoya0x00, line 109: I’m curious why did the build sometimes succeed without this line? 🤔 

I wonder if the builds of each dependency were running in parallel and `Component` was sometimes built before `Sponsor` was built by accident.

tomoya0x00 (approved):
LGTM 👍

app-ios/Modules/Package.swift
ry-itto: > Component was sometimes built before Sponsor was built by accident.

I think it's possible.
In last year's example, it was successful to build each module before building the app.

PR #756, opened 2023-08-19
#720 Highlight matched text

### LINKED ISSUE #720: Highlight searching text
**Idea Description**
I think it is useful!

S-H-Y-A:
## Issue
- close #720

## Overview (Required)
- Add hilight-matched-title feature.

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
Thanks! We are talking about the design. Please wait 🙏

takahirom:
@S-H-Y-A 
Thanks for challenging this implementation!
image
The designer suggest using pattern B. But if it is difficult can you implement B?

S-H-Y-A:
> @S-H-Y-A 
> Thanks for challenging this implementation!
> image
> The designer suggest using pattern B. But if it is difficult can you implement A?

Ok, I'll try to implement pattern B.

S-H-Y-A:
I' ve done it!
Look at The pictures below and give me some advice if you have.

## Screenshot
Light Theme | Dark Theme
:--: | :--:
 |

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableListItem.kt
```diff
@@ -55,6 +60,7 @@ fun TimetableListItem(
     onBookmarkClick: (TimetableItem) -> Unit,
     chipContent: @Composable RowScope.() -> Unit,
     modifier: Modifier = Modifier,
+    highlightIndexLange: (String) -> IntRange = { IntRange.EMPTY },
```
takahirom, line 63: How about having `searchingText: String? = null` instead of highlightIndexLange? I think it looks easier.

S-H-Y-A: I agree with you, I'll do coding again.

S-H-Y-A: Instead of `String?`, I choose `SearchListUiState`, which is value class of `String` and has some extension functions.

takahirom: I think you don't have to use `UiState` name for this. You can create `SearchQuery` or something. This is because we don't use UiState for components. And you can write some classes as usual composable classes in the official library.
https://github.com/DroidKaigi/conference-app-2023#component

S-H-Y-A: I changed from `highlightIndexLange: (String) -> IntRange = { IntRange.EMPTY },` to `highlightQuery: SearchQuery = SearchQuery.EMPTY,`.
How about it?

https://github.com/S-H-Y-A/conference-app-2023/blob/84b1ebd52b0aab801dccac2d8787e7f3fea85ac1/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableListItem.kt#L64

takahirom: There is a very little point. We don't use snake case for Compose code. 
https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#singletons-constants-sealed-class-and-enum-class-values

S-H-Y-A: Ops, that's my fault.
I have to read API guidelines again.
Thank you for your advice.

S-H-Y-A: I converted it into camel case and pushed.

takahirom (approved):
Looks great! We will have a great search experience!

takahirom:
@S-H-Y-A 

Could you take a look a test for merging this?

```
> Task :app-android:testDevDebugUnitTest

io.github.droidkaigi.confsched2023.KaigiAppTest > checkNavigateToStampsShot FAILED
    java.lang.NullPointerException at KaigiAppTest.kt:76

io.github.droidkaigi.confsched2023.KaigiAppTest > checkNavigateToBookmarkShot FAILED
    java.lang.NullPointerException at KaigiAppTest.kt:123

> Task :feature:sessions:testDebugUnitTest
```

S-H-Y-A:
It should work now!

takahirom:
> I found a fact that NullPointerExcption are thrown when value class is used as Default argument in Jetpack Compose
[This](https://github.com/S-H-Y-A/ValueClassTest) is an example project.

😮

takahirom (approved):
Looks great! Thanks

PR #692, opened 2023-08-18
Use filter chip for bookmark filter

kumokumot:
## Issue
- close #570 

## Overview (Required)
- replace the chip from AssistChip to FilterChip for bookmark filter
- changed to multiple FilterChips to be selected

## Links
- [Figma](https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?node-id=56145%3A47976)

## Screenshot
Before | After
:--: | :--:
 | 

https://github.com/DroidKaigi/conference-app-2023/assets/87587734/f5d2ebe2-53cc-49ee-a8d2-39b764b95578

Issue #570, opened 2023-08-16
Use FilterChip for bookmark filter


upon0426:
**Idea Description**
For more proper use of component, let's replace the chip from AssistChip to FilterChip

**Reference images and links**
<img width="400" alt="スクリーンショット 2023-08-16 22 39 46" src="https://github.com/DroidKaigi/conference-app-2023/assets/45708639/509b3efd-4352-41b8-9858-fbe445a47a47">

kumokumot:
🙋‍♂️can I try this issue?

satsukies:
@kumokumot Nice! Assigned 👍

PR #763, opened 2023-08-19
fix unnatural behavior of BookmarkFilterChip

kumokumot:
## Issue
- related to #570

## Overview (Required)
- fix the problem I put in https://github.com/DroidKaigi/conference-app-2023/pull/692 .
when Day1, Day2, and Day3 are all selected, the natural expectation is that tapping any of them will unselect the selected one, but in fact only the one you tapped will remain selected.

## Screenshot

Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/87587734/f7371850-ca5f-4409-8ae1-8c6b858dc32c" width="300" /> | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/87587734/f60037e4-8d1b-4341-9a23-1d411d462b07" width="300" />


takahirom (approved):
Thanks for fixing the bug! 🐛

PR #765, opened 2023-08-19
[iOS] Add Dynamic Tab Animation in TimetableDayHeader.

fumiyasac:
## Issue

https://github.com/DroidKaigi/conference-app-2023/issues/769

## Overview

In Android app, the date selection displayed in the header View element was accompanied by an animation.
But in the iOS app it didn't exist so I added it.

__【Summary】__

- Wrap the entire View element with `ZStack`.
- For the rectangular part of the background, the horizontal position and width are calculated using `GeometryReader`.
- Adjusted so that the background moves animation when switching dates.

## Links

Motivation: https://twitter.com/fumiyasac/status/1692857247829987802

## Screenshot

__【Movie】__

https://github.com/DroidKaigi/conference-app-2023/assets/949561/b8fc2231-a14d-4c1e-83d4-8e6527e24be5

__【iPhone】__

※ iPhone14 Pro Simulator

| Portrait |
| :--: |
|  |

| Landscape |
| :--: |
|  |

__【iPad】__

※ iPad mini (6th generation) Simulator

| Portrait |
| :--: |
|  |

| Landscape |
| :--: |
|  |

### DIFF
--- a/app-ios/Modules/Sources/Timetable/TimetableDayHeader.swift
+++ b/app-ios/Modules/Sources/Timetable/TimetableDayHeader.swift
@@ -15,9 +15,7 @@ struct TimetableDayHeader: View {
 
     // Define all button count to calculate holizontal position for capsule rectangle
     private var buttonsCount: Int {
-        [DroidKaigi2023Day]
-            .fromKotlinArray(DroidKaigi2023Day.values())
-            .count
+        Int(DroidKaigi2023Day.values().size)
     }
 
     var body: some View {
--- a/app-ios/Modules/Sources/Timetable/TimetableDayHeader.swift
+++ b/app-ios/Modules/Sources/Timetable/TimetableDayHeader.swift
@@ -69,16 +69,7 @@ struct TimetableDayHeader: View {
     }
 
     private func getIndexBySelectedDay() -> Int {
-        switch selectedDay {
-        case .day1:
-            return 0
-        case .day2:
-            return 1
-        case .day3:
-            return 2
-        default:
-            return 1
-        }
+        Int(selectedDay.ordinal)
     }
 
     private func calculateButtonWidth(deviceWidth: CGFloat) -> CGFloat {
--- a/app-ios/Modules/Sources/Timetable/TimetableDayHeader.swift
+++ b/app-ios/Modules/Sources/Timetable/TimetableDayHeader.swift
@@ -8,6 +8,18 @@ struct TimetableDayHeader: View {
     let selectedDay: DroidKaigi2023Day
     let onSelect: (DroidKaigi2023Day) -> Void
 
+    // Define margin values to calculate holizontal position for capsule rectangle
+    private let buttonAreaLeadingMargin = 16.0
+    private let buttonTrailingMargin = 16.0
+    private let betweenButtonMargin = 8.0
+
+    // Define all button count to calculate holizontal position for capsule rectangle
+    private var buttonsCount: Int {
+        [DroidKaigi2023Day]
+            .fromKotlinArray(DroidKaigi2023Day.values())
+            .count
+    }
+
     var body: some View {
         ZStack {
             AssetColors.Surface.surface.swiftUIColor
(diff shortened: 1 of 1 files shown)


app-ios/Modules/Sources/Timetable/TimetableDayHeader.swift
```diff
@@ -9,39 +9,62 @@ struct TimetableDayHeader: View {
...
-                    .clipShape(Capsule())
+                }
+            }
+            .padding(.horizontal, 16)
+            .padding(.vertical, 12)
+            .background(alignment: .center) {
+                GeometryReader { geometry in
+                    Capsule()
+                        .fill(AssetColors.Primary.primary.swiftUIColor)
+                        .frame(width: calcButtonWidth(width: geometry.size.width), height: 56)
```
tkhs0604, line 44: Thank you for your first PR to this repo! Looks awesome to me 🤩

I have 1 request.
If possible, can you define some constants used in this component to make it easier to understand what are these magic numbers? 🙏

fumiyasac: @tkhs0604 

Thank you for request:bow:
The following values ​​are related to calculations, so they are defined separately.

```swift
// Define margin values to calculate holizontal position for capsule rectangle
private let buttonAreaLeadingMargin = 16.0
private let buttonTrailingMargin = 16.0
private let betweenButtonMargin = 8.0

// Define all button count to calculate holizontal position for capsule rectangle
private var buttonsCount: Int {
    [DroidKaigi2023Day]
        .fromKotlinArray(DroidKaigi2023Day.values())
        .count
}
```

By the way, refactor the methods related to computation.

tkhs0604: Thank you, they are so readable! Let me check again 🙆‍♂️

app-ios/Modules/Sources/Timetable/TimetableDayHeader.swift
```diff
@@ -8,40 +8,88 @@ struct TimetableDayHeader: View {
...
+        switch selectedDay {
+        case .day1:
+            return 0
+        case .day2:
+            return 1
+        case .day3:
+            return 2
+        default:
+            return 1
         }
```
ry-itto, line 81: How about use `ordinal` property? 👀 
It might be the same as current one.

app-ios/Modules/Sources/Timetable/TimetableDayHeader.swift
```diff
@@ -8,40 +8,88 @@ struct TimetableDayHeader: View {
...
+    // Define margin values to calculate holizontal position for capsule rectangle
+    private let buttonAreaLeadingMargin = 16.0
+    private let buttonTrailingMargin = 16.0
+    private let betweenButtonMargin = 8.0
+
+    // Define all button count to calculate holizontal position for capsule rectangle
+    private var buttonsCount: Int {
+        [DroidKaigi2023Day]
+            .fromKotlinArray(DroidKaigi2023Day.values())
+            .count
```
ry-itto, line 20: Maybe `DroidKaigi2023Day.values().size` is useful 👀

ry-itto (commented):
Good catch and so cool! Thank you!
I left some review points, please check that!

app-ios/Modules/Sources/Timetable/TimetableDayHeader.swift
tkhs0604: Oh, this is exactly what I wanted to check if we can use or not! 👍

fumiyasac: @ry-itto 

Thank you for review! I try to use `ordinal` property.
It seems to fix `getIndexBySelectedDay()` as below👍 

```swift
private func getIndexBySelectedDay() -> Int {
    Int(selectedDay.ordinal)
}
```

tkhs0604 (approved):
@fumiyasac 
LGTM from my side! Thank you for providing the same user experience as Android 🤖

app-ios/Modules/Sources/Timetable/TimetableDayHeader.swift
fumiyasac: @ry-itto

Thank you for review!
It seems to fix `buttonsCount` property as below👍

```
private var buttonsCount: Int {
    Int(DroidKaigi2023Day.values().size)
}
```

fumiyasac:
@tkhs0604 
Thank you for review ! 
I have adjusted the content of the review I received from @ry-itto earlier.

ry-itto (approved):
Looks great! Thank you for improving!

PR #772, opened 2023-08-19
correct About screen title color

### LINKED ISSUE #787: Color of AboutScreen TopAppBar title is difficult to see when in dark mode
**Describe the bug**
The color of the title of the TopAppBar in the AboutScreen is difficult to see when scrolling in dark mode

Perhaps the color of the following text.
The color of text when not scrolling is not defined, so it is not necessary to define a color.
https://github.com/DroidKaigi/conference-app-2023/blob/29922658fc76a516fec71ca2f498d021dc9d6dfb/feature/about/src/main/java/io/github/droidkaigi/confsched2023/about/AboutScreen.kt#L110-L114

**To Reproduce**
1. set to dark mode
2. go to AboutScreen
3. scroll down

**Expected behavior**
Titles are not the same color as the background, even in dark mode, and are easily visible.

**Screenshots**
|Dark mode|Light mode|
|---|---|
|<video src="https://github.com/DroidKaigi/conference-app-2023/assets/60963155/7019a37e-34c1-4ca4-a985-55ab3797a6b5">|<video src="https://github.com/DroidKaigi/conference-app-2023/assets/60963155/106c85f6-b4be-43ef-93a3-8e255c57d016">|

ked4ma:
## Issue
- close #787

## Overview (Required)
- title color was displayed incorrectly on about screen when it is dark mode
- enable to display with the title color for each mode (light / dark)

## Links
- [figma](https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=54462-36316&mode=design&t=2cVHyvhRdMdQbkGU-0)

## Screenshot
Before | After
:--: | :--:
<br><br><br> | <br><br><br>

## records
| Before  | After |
| ------------- | ------------- |
| <video src="https://github.com/DroidKaigi/conference-app-2023/assets/74723074/a0ce30c3-0700-4a24-a40c-7e73b57e99da"> | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/74723074/afd377b4-9b0c-4ebc-bbcc-4933e9e5d84a"/> |
| <video src="https://github.com/DroidKaigi/conference-app-2023/assets/74723074/10b32cab-4002-4d09-8065-2d0529348704" /> | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/74723074/f020ab6c-1bd2-4c29-b611-27276df3f7a2"/> |

### DIFF
--- a/feature/about/src/main/java/io/github/droidkaigi/confsched2023/about/AboutScreen.kt
+++ b/feature/about/src/main/java/io/github/droidkaigi/confsched2023/about/AboutScreen.kt
@@ -21,7 +21,7 @@ import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.draw.alpha
 import androidx.compose.ui.input.nestedscroll.nestedScroll
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.LocalLayoutDirection
@@ -109,8 +109,8 @@ private fun AboutScreen(
                     } else {
                         Text(
                             text = AboutStrings.Title.asString(),
-                            color = Color.Unspecified.copy(alpha = scrollBehavior.state.overlappedFraction),
                             style = MaterialTheme.typography.titleLarge,
+                            modifier = Modifier.alpha(scrollBehavior.state.overlappedFraction),
                         )
                     }
                 },


feature/about/src/main/java/io/github/droidkaigi/confsched2023/about/AboutScreen.kt
```diff
@@ -109,8 +109,8 @@ private fun AboutScreen(
                     } else {
                         Text(
                             text = AboutStrings.Title.asString(),
-                            color = Color.Unspecified.copy(alpha = scrollBehavior.state.overlappedFraction),
```
ked4ma, line 112: Text will decide with following code.
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/Text.kt;l=112-116?hl=ja
so, this copy of `Color.Unspecified` is not used.
Of course, we can specify color as `LocalContentColor.current.copy(alpha = scrollBehavior.state.overlappedFraction)` (refer to Text.kt)
but I used `Modifier.alpha(...)` because of code simplicity.
(if set Modifier, we don't have to care about COLOR of text, I considered)

Please comment if specifying color is better. (in that case, I'll change code)

ked4ma:
Differences of snapshot are occuerred because `alpha` is reflected by this change.
(Before this pr, color arg was not used)

feature/about/src/main/java/io/github/droidkaigi/confsched2023/about/AboutScreen.kt
takahirom: That makes sense!

takahirom:
Thank you for considering how we improve the code!

Issue #383, opened 2023-08-07
Make the time label stickable on timetable screen


tkhs0604:
## What is required
It is better for the time label on timetable screen to be stickable when scrolling.
<img src="https://github.com/DroidKaigi/conference-app-2023-images/assets/67794229/19bf1779-6743-4875-914d-b4f25c735332" width="300">

## Reference
https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=54902-43788&mode=design&t=byz4fuqUd6q4I08I-0

KwakEuiJin:
I want to resolve this issue, can i take the issue?

takahirom:
Thanks! Assigned!

KwakEuiJin:
I'm working on this issue. It is a rather difficult problem, so it will take some time 🥲 but i keep trying

KwakEuiJin:
https://github.com/DroidKaigi/conference-app-2023/assets/93872496/3906325b-b108-4484-bd88-f605e44b2c55

@takahirom 
@tkhs0604 

Please check if this is a good way

However, in the case of the last index, the time is not visible. We'll keep looking for a solution

tkhs0604:
@KwakEuiJin 
I completely missed your comment. I really apologize 🙇‍♂️

Thank you so much for your investigation and trial implementation.
The following comments are just my impression:
- Looks much better than the current implementation
- The behavior does not look a general sticky header, but it's acceptable for me.
- I would like the last duration information to be displayed (I think it is one of this issue's requirements 😢)

FYI: This may be similar to what we want.
https://fvilarino.medium.com/creating-a-sticky-letter-list-in-jetpack-compose-2e7e9c8f9b91

@takahirom 
Do you have any comments about the above movie? 👀

S-H-Y-A:
I didn't realize this issue and @KwakEuiJin has been trying to implement that.
I already made it. ( Look → #775 )
If you have any advice to me, give me some one!

KwakEuiJin:
@S-H-Y-A 
I applaud the implementation in a better way. I will look at your code and study more 👍

@tkhs0604 
Please change assigned, thanks

S-H-Y-A:
#775 was marged.
I close this issue.

takahirom:
Thanks!

PR #775, opened 2023-08-19
[Idea] Make timeText of ScheduleScreen sticky

S-H-Y-A:
## Issue
- ~~None~~
- #383

## Overview (Required)

- Make timeText of ScheduleScreen sticky.
  - ScheduleScreen, SearchSection and BookmarkSection are modified.
  - We don't have to scroll to check when a speech is held.

### Before
https://github.com/DroidKaigi/conference-app-2023/assets/74596628/269599b9-1a1d-435b-80ad-536e1def062b

### After
https://github.com/DroidKaigi/conference-app-2023/assets/74596628/cfc20d99-f947-4f1b-8853-9ab550826854


tkhs0604:
Wow, awesome! ✨
Maybe it is related to https://github.com/DroidKaigi/conference-app-2023/issues/383 👀

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/section/BookmarkList.kt
```diff
@@ -36,65 +45,90 @@ fun BookmarkList(
...
 ) {
+    val density = LocalDensity.current
     LazyColumn(
         state = scrollState,
         modifier = modifier.padding(end = 16.dp),
     ) {
-        timetableItemMap.forEach { (_, values) ->
-            itemsIndexed(values) { index, timetableItem ->
-                Row(modifier = Modifier.padding(top = 10.dp)) {
+        itemsIndexed(timetableItemMap.toList(), key = { _, (key, _) -> key }) { index, (_, values) ->
```
takahirom, line 53: We might have to consider that the item for LazyColumn is going to be a time range.
So even if we don't see all the items for the time range, it will be created.
On the other hand, if we see the scroll state and make other LazyColumn for time, we don't have this issue.
I'm not sure if this problem is noticeable 🤔

S-H-Y-A: Um..., the number of timeTable item which an item of LazyColumn is 4 or five at most. Size of 4 item is the same as that of Screen. I think it is not a big problem.
The size of item Of LazyColumn are different from each other, so it is difficult to calclate those each height and apply Offset to TimeText.

takahirom: Yeah, it works awesome and codes are great so if we need to improve I think we can. I'm willing to merge this.

takahirom:
Great work! Thanks!

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/section/BookmarkList.kt
S-H-Y-A: I'll mark this as solved.

PR #781, opened 2023-08-20
modify session grid speaker text to fix bottom

### LINKED ISSUE #731: Fix the speaker to the bottom.
**Idea Description**
Currently, the session grid items are rendered as shown in the image below.
We would like the speaker to be fixed on the bottom to match the Figma design.

Figma | Actual
 :-: | :-:
 |

Aniokrait:
## Issue
- close #731

## Overview (Required)
- fix the problem that a session speaker text was not fixed to bottom like in Figma.

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableGridItem.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableGridItem.kt
@@ -127,7 +127,7 @@ fun TimetableGridItem(
             Spacer(
                 modifier = Modifier
                     .weight(1f)
-                    .defaultMinSize(minHeight = 8.dp)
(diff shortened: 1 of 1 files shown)


feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableGridItem.kt
```diff
@@ -123,7 +124,11 @@ fun TimetableGridItem(
                 )
             }
 
-            Spacer(modifier = Modifier.height(TimetableGridItemSizes.scheduleToSpeakerSpaceHeight))
+            Spacer(
+                modifier = Modifier
+                    .weight(1f)
```
takahirom, line 129: 👍

takahirom (approved):
Looks great! Thank you for your simple implementation!

PR #782, opened 2023-08-20
Create AppVersion for provide BuildConfig to any module

### LINKED ISSUE #628: Improved method of retrieving version names
**Idea Description**
I need to fetch the version name every time a Composable function is re-drawn, and I'd like to improve performance in some way. I'd appreciate any good ideas you might have (e.g., using the remember{} method).

https://github.com/DroidKaigi/conference-app-2023/blob/efc6880ff10f3aa27a7037e803b899730a0290c5/feature/about/src/main/java/io/github/droidkaigi/confsched2023/about/AboutScreen.kt#L65

matsumo0922:
## Issue
- closes #628 

## Overview (Required)
- Create `AppVersion` data class for provide BuildConfig to any module
  - It is singlton so better performance than before implementation
- Remove useless implementation

### DIFF
--- a/app-android/src/main/java/io/github/droidkaigi/confsched2023/AppModule.kt
+++ b/app-android/src/main/java/io/github/droidkaigi/confsched2023/AppModule.kt
@@ -1,0 +1,24 @@
+package io.github.droidkaigi.confsched2023
+
+import dagger.Module
+import dagger.Provides
+import dagger.hilt.InstallIn
+import dagger.hilt.components.SingletonComponent
+import io.github.droidkaigi.confsched2023.model.AppVersion
+import javax.inject.Singleton
+
+@Module
+@InstallIn(SingletonComponent::class)
+object AppModule {
+
+    @Provides
+    @Singleton
+    fun provideAppVersion(): AppVersion {
+        return AppVersion(
+            packageName = BuildConfig.APPLICATION_ID,
+            versionName = BuildConfig.VERSION_NAME,
+            versionCode = BuildConfig.VERSION_CODE,
+            isDebug = BuildConfig.DEBUG,
+        )
+    }
+}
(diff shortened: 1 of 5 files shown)


takahirom (approved):
I didn't think of this! Thanks!

takahirom:
We have to fix the test. Can you take a look?

matsumo0922:
I am having trouble resolving the Dagger/Hilt error. When run the test, it can't seem to do the `AppVersion` inject because it says `MissingBinding`.

I have tried several different approaches but have not been able to resolve it. 
Does anyone have a good solution?

takahirom:
I think we can fix this by adding these settings into `core/testing build.gradle.kts`. But I think you shouldn't do it because the core module shouldn't access the application module. It's very difficult. Maybe we should create a module that have variant and dagger module :sob:

```
android.defaultConfig {
    missingDimensionStrategy("network", "dev")
}

dependencies {
    implementation(project(":app-android"))
```

takahirom:
Well, using `remember{}` is one option. 👀

matsumo0922:
Uhmm...
I thought this approach was generic and interesting, but if it seems difficult to modify the test, it may be easier to use `remember`.

I'll worry about it some more and if it doesn't seem to solve the problem, I'll close this PR and create a PR that uses `remember`.

takahirom:
Now I see

floormap -> testImplementation(projects.core.testing) → AboutViewModel 😭

takahirom:
One option might be placing the module in core/common and using PackageManager or (app as AppVersionHolder)?.getAppVersion()

matsumo0922:
@takahirom 

Mocked AppVersion in all tests bf8aa08

This is not a good solution, and considering the original issue #628, we don't have to go this far, just use `remember`. However, this approach is generic and interesting, and I would like to have it merged as is, if only takahirom san is willing to do so. :bow:

takahirom:
Well, we can have the binding for each test but I don't think it is a practical solution. If we want to change something like a server endpoint URL. Will we add the URL to every test again..?

matsumo0922:
Isn't it...

The approach here may have been a little too difficult. The problem isn't worth the effort, and I can't keep working on it forever, so I'm going to switch to using remember here. (FYI: #880)

However, I have never given up, and I would like to issue a PR if a better approach is found.

PR #785, opened 2023-08-20
Remove duplicated dependencies

sys1yagi:
## Issue

none

## Overview (Required)

I learned that the build-logic module can declare a set of settings that can be commonly used in a project as a plugin.
Then, realized that I was declaring duplicate dependencies in the modules that use those plugins.
So I removed the duplicate declarations in the various modules.

Question: It appears that `libs.composeHiltNavigtation` and `libs.composeMaterialIcon` declare a dependency in all modules that use `androidfeature` plugin. Should these be included in AndroidComposePlugin?

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/about/build.gradle.kts
+++ b/feature/about/build.gradle.kts
@@ -10,15 +10,6 @@ dependencies {
     implementation(projects.core.model)
     testImplementation(projects.core.testing)
 
-    implementation(libs.androidxCoreKtx)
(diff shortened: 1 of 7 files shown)


takahirom (approved):
Welcome back to DroidKaigi App! You've removed 63 lines! Thanks!

Issue #636, opened 2023-08-17
Connecting iOS and Android with Kotlin


takahirom:
**Idea Description**
Is anyone interested in working on connecting iOS and Android using Kotlin?

Objective: Enable the use of implementations found in data/commonMain, such as invoking APIs on iOS.

Currently, iOS requires a setup like ApiModule, so it will likely involve setting up components like Ktor, DataStore, etc. Integration of a DI (Dependency Injection) framework is an option, or using a singleton object such as object ApiModule { val ktor = ... } is also considered acceptable.

**Reference images and links**
https://github.com/DroidKaigi/conference-app-2023/blob/192521ac03a3c5c499a5019bfc49be2094a36eeb/core/data/src/androidMain/kotlin/io/github/droidkaigi/confsched2023/data/di/ApiModule.kt#L31

takahirom:
I tried this

PR #791, opened 2023-08-20
Add Koin and Add KMP binding for iOS

takahirom:
## Issue
- close https://github.com/DroidKaigi/conference-app-2023/issues/636

## Overview (Required)
- 

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |


build-logic/src/main/kotlin/io/github/droidkaigi/confsched2023/primitive/KmpIosPlugin.kt
```diff
@@ -18,13 +18,27 @@ class KmpIosPlugin : Plugin<Project> {
             }
 
             kotlin {
+                // https://github.com/JetBrains/compose-multiplatform/issues/3135#issuecomment-1655877617
```
takahirom, line 21: The test fails if I don't specify these options.

takahirom, line 21:
```suggestion
                // TODO: remove after 1.5 release
                //  https://github.com/JetBrains/compose-multiplatform/issues/3135#issuecomment-1655877617
```

core/data/build.gradle.kts
```diff
@@ -20,7 +20,6 @@ kotlin {
                 implementation(libs.kotlinxCoroutinesCore)
                 // We use api for test
                 api(libs.androidxDatastoreDatastorePreferences)
-                implementation(libs.multiplatformFirebaseAuth)
```
takahirom, line 23: This causes build error that says missing binding

PR #799, opened 2023-08-20
[iOS] Add missing 'Component' dependency to Sponsor target

ShoMasegi:
## Issue
Nothing

## Overview (Required)
- add missing dependency (`Component`) to `Sponsor` target

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |


tkhs0604 (approved):
Good catch! Thank you for the fix! 🛠

PR #804, opened 2023-08-20
Stores `viewModelScope` in `KmpViewModel` on iOS

kubode:
## Issue
- No issue

## Overview (Required)
- I found that `KmpViewModel.viewModelScope` is a computed property on iOS, making a new instance per accessing to `viewModelScope`. Since the scope is different each time it is accessed, it causes unintended behavior.
- The TODO comment has been left because I couldn't find a solution.


takahirom (approved):
Making scope for each time is definitely a bad idea! Thanks for fixing this!

PR #805, opened 2023-08-20
Use `@OptionalExpectation` instead

kubode:
## Issue
- No issue

## Overview (Required)
- Kotlin Multiplatform has an experimental [`@OptionalExpectation`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-optional-expectation/) annotation that makes the need not to implement the `expect` annotation on all platforms.


takahirom (approved):
I didn't know this! Awesome!

PR #807, opened 2023-08-20
Use `Locale.current` for `lang()`

kubode:
## Issue
- No issue

## Overview (Required)
- Compose Multiplatform has `Locale` mostly like Android's one. I think it's better than implementing with expect/actual.
image


tkhs0604 (approved):
Wow, I didn't know that it can be used in common module!
Thank you for simplifying multilanguage implementation! 🇺🇸🇯🇵

PR #808, opened 2023-08-20
Fix SearchList padding

k-shinn:
## Issue
- none

## Overview (Required)
- Fixed `SearchList` padding based on Figma's design
  - before : Set Padding to `end`
  - after : Set Padding to `start`

## Links
- Figma SearchList
  - https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=58036-78614&mode=dev

## Screenshot
Before | After
:--: | :--:
|
|


takahirom (approved):
Looks great!

PR #811, opened 2023-08-20
:wrench: Replace lambda function to callable reference.

### LINKED ISSUE #757: [Idea]: Replace lambda function to callable reference
**Idea Description**

Why we should use callable reference instead of lambda function?
we don't wanna care about composition, but lambda function instantiate which has not same hash. 
`Composable` will compose when hashes of args is different, so `Composable` with lambda instance has many point to `re`compose.
callable reference has same hash when instance is equal, so it is good to jetpack compose user.

```kotlin
at Composable
fun Hoge(onClickFoo: () -> Unit)
```

before
```kotlin
Hoge(
    onClickFoo = {
        navController.navigateToBar()
    }
)
```

after
```kotlin
Hoge(
    onClickFoo = navController::navigateToBar
)
```

https://kotlinlang.org/docs/reflection.html#callable-references

Corvus400:
## Issue
- close #757 

## Overview (Required)
- As the title says.
- I looked around in the project and couldn't find anything that could be replaced other than KaigiApp.kt, so I'm wondering if this is really the right thing to do. 🤔

### DIFF
--- a/app-android/src/main/java/io/github/droidkaigi/confsched2023/KaigiApp.kt
+++ b/app-android/src/main/java/io/github/droidkaigi/confsched2023/KaigiApp.kt
@@ -114,42 +114,24 @@ private fun KaigiNavHost(
     NavHostWithSharedAxisX(navController = navController, startDestination = mainScreenRoute) {
         mainScreen(windowSize, displayFeatures, navController, externalNavController)
         sessionScreens(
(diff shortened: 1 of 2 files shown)


takahirom (approved):
Looks great! Thank you for making the app fast!
(I don't know why screeshot is comparing with old image🤔)

PR #812, opened 2023-08-20
:recycle: Add template text for when pasting videos.

Corvus400:
## Issue
- none.

## Overview (Required)
- I added the template because I wanted to have before/after display when pasting videos as well.
- The video below is a sample of the display.

## Movie(PR Template sample)
Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/f9a89c8f-dccd-4505-96d8-5b1b38b51616" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/a3d6f7c9-68a8-4ed7-89c4-da9c076935df" width="300" >

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,3 +11,8 @@
 Before | After
 :--: | :--:
 <img src="" width="300" /> | <img src="" width="300" />
+
+## Movie
+Before | After
+:--: | :--:
+<video src="" width="300" > | <video src="" width="300" >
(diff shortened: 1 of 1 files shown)


takahirom:
Thanks. We can easily confirm the behavior with this but I think there is a tradeoff that beginners may find challenging. 👀

Corvus400:
I was thinking of this as an optional template for attaching images or videos.
Should I add a comment out indicating that it is not required?🙇‍♂️

takahirom:
That would be great!

takahirom:
I was wanting to encourage the screenshot test. So I added it as well.

takahirom:
Thanks for bringing this to my attention!

PR #847, opened 2023-08-21
chore: update bundler

RyuNen344:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- 

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,10 +2,7 @@
 
 source "https://rubygems.org"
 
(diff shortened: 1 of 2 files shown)


tkhs0604 (approved):
The sooner, the better! Thank you for introducing a brand-new feature!

PR #852, opened 2023-08-21
[iOS] update timetable list item view

tkhs0604:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- The design of timetable list item was updated

## Links
- https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?node-id=56145%3A49549&mode=dev

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/app-ios/Modules/Sources/Timetable/InfoLabel.swift
+++ b/app-ios/Modules/Sources/Timetable/InfoLabel.swift
@@ -26,10 +26,10 @@ struct InfoLabel: View {
             .padding(.vertical, 4)
             .foregroundColor(labelColor)
             .background(backgroundColor)
-            .cornerRadius(4)
+            .cornerRadius(50)
             .overlay {
                 if let strokeColor = strokeColor {
-                    RoundedRectangle(cornerRadius: 4)
+                    RoundedRectangle(cornerRadius: 50)
                         .inset(by: 0.5)
                         .stroke(strokeColor, lineWidth: 1)
                 } else {
--- a/app-ios/Modules/Sources/Timetable/TimetableListItemView.swift
+++ b/app-ios/Modules/Sources/Timetable/TimetableListItemView.swift
@@ -5,11 +5,11 @@ import Theme
 
 struct TimetableListItemView: View {
     let timetableItemWithFavorite: TimetableItemWithFavorite
-    
+
     var timetableItem: TimetableItem {
         self.timetableItemWithFavorite.timetableItem
     }
-    
+
     var body: some View {
         HStack(alignment: .top) {
             VStack(alignment: .leading, spacing: 0) {
--- a/app-ios/Modules/Sources/Timetable/TimetableListItemView.swift
+++ b/app-ios/Modules/Sources/Timetable/TimetableListItemView.swift
@@ -5,17 +5,17 @@ import Theme
 
 struct TimetableListItemView: View {
     let timetableItemWithFavorite: TimetableItemWithFavorite
-
+    
     var timetableItem: TimetableItem {
         self.timetableItemWithFavorite.timetableItem
     }
-
+    
     var body: some View {
         HStack(alignment: .top) {
             VStack(alignment: .leading, spacing: 0) {
-                Spacer().frame(height: 4)
+                Spacer().frame(height: 16)
                 // TODO apply like flexbox layout
-                LazyHStack(spacing: 4) {
+                HStack(spacing: 4) {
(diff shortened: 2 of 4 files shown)


app-ios/Modules/Sources/Timetable/TimetableListItemView.swift
```diff
@@ -5,17 +5,17 @@ import Theme
...
-
+    
     var body: some View {
         HStack(alignment: .top) {
             VStack(alignment: .leading, spacing: 0) {
-                Spacer().frame(height: 4)
+                Spacer().frame(height: 16)
                 // TODO apply like flexbox layout
-                LazyHStack(spacing: 4) {
+                HStack(spacing: 4) {
```
tkhs0604, line 18: If LazyHStack is used, the title text is displayed but only 1 line, so changed it to HStack.

app-ios/Modules/Sources/Timetable/TimetableListItemView.swift
```diff
@@ -29,12 +29,11 @@ struct TimetableListItemView: View {
                         )
                     }
                 }
-                Spacer().frame(height: 12)
+                Spacer().frame(height: 8)
                 Text(timetableItem.title.currentLangTitle)
                     .multilineTextAlignment(.leading)
                     .font(Font.system(size: 22, weight: .medium, design: .default))
                     .foregroundStyle(AssetColors.Surface.onSurface.swiftUIColor)
-                    .frame(maxWidth: .infinity, alignment: .leading)
```
tkhs0604, line 37: Looks unnecessary.

app-ios/Modules/Sources/Timetable/InfoLabel.swift
```diff
@@ -26,10 +26,10 @@ struct InfoLabel: View {
             .padding(.vertical, 4)
             .foregroundColor(labelColor)
             .background(backgroundColor)
-            .cornerRadius(4)
+            .cornerRadius(50)
             .overlay {
                 if let strokeColor = strokeColor {
-                    RoundedRectangle(cornerRadius: 4)
+                    RoundedRectangle(cornerRadius: 50)
```
ry-itto, line 32: How about using [capsule](https://developer.apple.com/documentation/swiftui/capsule) ?
I think this is useful for this situation!

app-ios/Modules/Sources/Timetable/TimetableListItemView.swift
```diff
@@ -5,17 +5,17 @@ import Theme
 
 struct TimetableListItemView: View {
     let timetableItemWithFavorite: TimetableItemWithFavorite
-
+
```
ry-itto, line 8: I think this causes SwiftLint warning 👀 
Please check this 🙏🏼

app-ios/Modules/Sources/Timetable/InfoLabel.swift
tkhs0604: Thank you, it looks much better 👍 

`Capsule` has the same look as `RoundedRectangle(cornerRadius: 50)`, but actually I could not understand why this one works well...
The tag component's height is 24 and its required radius is 50, which is much larger than the half size of its height.
Maybe, the cornerRadius of RoundedRectangle was not respected...? 

RoundedRectangle(cornerRadius: 50) | Capsule
:--:|:--:
 |

app-ios/Modules/Sources/Timetable/TimetableListItemView.swift
tkhs0604: Ah, that's true. I will fix it 🙏
By the way, my Xcode arranges my code like this line by using `control + i` command ☹️
Is there any good way to avoid this incorrect formatting by Xcode?

ry-itto: > Is there any good way to avoid this incorrect formatting by Xcode?

This setting is useful! `Automatically trim trailing whitespace`

スクリーンショット 2023-08-23 2 59 45

tkhs0604: Thank you! Reformatted! https://github.com/DroidKaigi/conference-app-2023/pull/852/commits/7012ea6c90cced19c569c982ba45afe8eb8085db

ry-itto (approved):
LGTM! 👍🏼

PR #855, opened 2023-08-21
chore: run ios unit test on ci

RyuNen344:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- 

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


RyuNen344:
fastlane scan always hangout on ci.... what went wrong.....

PR #859, opened 2023-08-22
Select current day on initial view of timetable

katsu-gen:
## Issue
- none

## Overview (Required)
- The most recent or current day (After the event,  ~~last~~first day) would be selected in initial view of timetable.

## Links
- none

## Movie
Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/35623289/12ce89b8-622e-4e89-886a-01700f302f2c" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/35623289/da5779c4-d627-4141-be77-91232ba1aae8" width="300" >


core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2023/model/DroidKaigi2023Day.kt
```diff
@@ -81,5 +81,17 @@ public enum class DroidKaigi2023Day(
         public fun defaultDyamicThemeDate(): Boolean {
             return Day1.start < Clock.System.now()
         }
+
+        /**
+         * @return appropriate day at this time
+         */
+        fun currentDay(): DroidKaigi2023Day {
+            lateinit var currentDay: DroidKaigi2023Day
+            for (entry in entries.sortedBy { it.day }) {
```
katsu-gen, line 90: Sort once so that it can withstand cases where the definition order is different from the order of event day.

tkhs0604:
@katsu-gen 
Thank you for your proposal! Looks very convenient. Let me check 👀

core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2023/model/DroidKaigi2023Day.kt
```diff
@@ -81,5 +81,17 @@ public enum class DroidKaigi2023Day(
...
+                currentDay = entry
+                if (Clock.System.now() <= entry.end) break
```
tkhs0604, line 92: We would like to display Day1 tab after the conference ended.
Would you modify the logic? 🙏

tkhs0604 (commented):
Thank you for improving timetable screen! ✨
I left 1 request and 1 nitpicking, so I would like you to check when you have time 🙏

core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2023/model/DroidKaigi2023Day.kt
```diff
@@ -81,5 +81,17 @@ public enum class DroidKaigi2023Day(
         public fun defaultDyamicThemeDate(): Boolean {
             return Day1.start < Clock.System.now()
         }
+
+        /**
+         * @return appropriate day at this time
+         */
+        fun currentDay(): DroidKaigi2023Day {
```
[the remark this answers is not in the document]
katsu-gen: Thank you for better suggestion! I applied it.

core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2023/model/DroidKaigi2023Day.kt
katsu-gen: Thank you for comment. I modified the logic to select Day1 when it is after the conference ended.

core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2023/model/DroidKaigi2023Day.kt
```diff
@@ -81,5 +81,17 @@ public enum class DroidKaigi2023Day(
         public fun defaultDyamicThemeDate(): Boolean {
             return Day1.start < Clock.System.now()
         }
+
+        /**
+         * @return appropriate initial day for now
+         */
+        fun initialSelectedDay(): DroidKaigi2023Day {
+            val reversedEntries = entries.sortedByDescending { it.day }
```
tkhs0604, line 89: Thank you for refactoring! It worked as we expected! 👍

By the way, I came up with another solution:
```kotlin
fun initialSelectedDay(): DroidKaigi2023Day {
    val now = Clock.System.now()
    return ofOrNull(now) ?: Day1
}
```
The above one is not so accurate as your code since my code does not consider start and end time.
On the other hand, it looks a bit simpler 💡

tkhs0604: But your idea and implementation are so good 👍 I marked it as approved 💯

tkhs0604 (approved):
Thank you for improving it again!

PR #865, opened 2023-08-22
[iOS] Connect API using KMP module

ry-itto:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- 

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


tkhs0604 (approved):
❤️

ry-itto:
> I was wondering if you could try this. But I'll leave it to you 🙇 https://github.com/rickclephas/KMP-NativeCoroutines#swift-swift-package-manager #862

I think introducing KMP-NativeCoroutines is not only in iOS, so I didn't in this PR. But I'll try it!

PR #867, opened 2023-08-22
Fix time text format

### LINKED ISSUE #834: Color of text is wrong in timetable screen
**Idea Description**
The session end time should have different color

スクリーンショット 2023-08-21 22 03 04

CookieySun:
## Issue
- close #834

## Overview (Required)
- Add color and font weight to timetable.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/section/TimetableList.kt
```diff
@@ -94,9 +96,13 @@ fun TimetableList(
                             text = timetableItem.startsTimeString,
                             fontWeight = FontWeight.Medium,
                         )
-                        Text(text = "|")
+                        Text(text = "|",
+                            color = md_theme_light_outlineVariant,
+                        )
                         Text(
                             text = timetableItem.endsTimeString,
+                            color = md_theme_light_secondary,
```
takahirom, line 104: I think we should use a Theme color for this because it will break when you use dark theme 👀

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/section/TimetableList.kt
```diff
@@ -94,9 +96,13 @@ fun TimetableList(
                             text = timetableItem.startsTimeString,
                             fontWeight = FontWeight.Medium,
                         )
-                        Text(text = "|")
+                        Text(text = "|",
+                            color = md_theme_light_outlineVariant,
+                        )
```
takahirom, line 101: Could you put this back?  This is because there is a different PR that will cause conflict 🙏

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/section/TimetableList.kt
CookieySun: I fix it!

takahirom (approved):
Looks better! Thanks!

PR #868, opened 2023-08-22
Add BookmarkScreenTest

itochan:
## Issue
None

## Overview (Required)
- Add BookmarkTest and BookmarkScreenRobot

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):
Looks great! Thanks for creating new test!🔖

PR #877, opened 2023-08-22
Replace Chip Design with Newly Defined Tag

### LINKED ISSUE #826: Replace `Chip` Design with Newly Defined `Tag`
**Idea Description**
Currently, the Chip design is used in the application, and it comes with a constraint of having a larger tap target, preventing it from being resized smaller. To address this and embrace a new appearance with rounded corners, we need to replace the existing Chip design with the newly defined Tag. This change will allow for a more flexible and updated look within the application.

image

https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?node-id=54462%3A36316&mode=dev

Hachimori:
## Issue
- close #826 

## Overview (Required)
- Changed Timetable item chip's outline into rounded corner.

## Links
- 

## Screenshot

- Timetable

Before | After
:--: | :--:
 | 

- Bookmark

Before | After
:--: | :--:
 | 

- Search

Before | After
:--: | :--:
 |


tkhs0604:
@Hachimori 
I apologize for misleading you 🙇‍♂️
My last comment means that the height of the room chips need to be adjusted.

And I found that [the chip components had 32dp height](https://m3.material.io/components/chips/specs) and it could not be changed. 
On the other hand, the new design requires 24dp height. 
Thus, I think we need to create a new tag composable and replace the existing chips with it to resolve this diff.

Again, I am really sorry 🙏

Hachimori:
@tkhs0604

Thank you for the explanation, I understand the requirement!

I made following changes:
- Created `SessionTag` whose height is 24dp. (I named it `SessionTag` since just `Tag` is a bit ambiguous.)
- Replaced existing `*Chip`s with `SessionTag`.

Please kindly review the code again 🙇‍♂️

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/SessionTag.kt
```diff
@@ -0,0 +1,66 @@
...
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import io.github.droidkaigi.confsched2023.designsystem.preview.MultiThemePreviews
+import io.github.droidkaigi.confsched2023.designsystem.theme.KaigiTheme
+import io.github.droidkaigi.confsched2023.designsystem.theme.hallColors
+
+@Composable
+fun SessionTag(
```
tkhs0604, line 23: > Created SessionTag whose height is 24dp. (I named it SessionTag since just Tag is a bit ambiguous.)

❤️
It does much make sense 👍

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/SessionTag.kt
```diff
@@ -0,0 +1,66 @@
...
+    label: String,
+    modifier: Modifier = Modifier,
+    labelColor: Color = MaterialTheme.colorScheme.onSurfaceVariant,
+    backgroundColor: Color = Color.Transparent,
+    borderColor: Color = MaterialTheme.colorScheme.outline,
```
tkhs0604, line 28: [IMO]
It might be better to change its type to nullable with default null value, and add border modifier conditionally.

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableListItem.kt
```diff
@@ -71,7 +72,11 @@ fun TimetableListItem(
             .clickable { onClick(timetableItem) },
     ) {
         Row(verticalAlignment = Alignment.CenterVertically) {
-            FlowRow(modifier = Modifier.weight(1F)) {
+            FlowRow(
+                modifier = Modifier.weight(1F),
+                horizontalArrangement = Arrangement.spacedBy(4.dp),
```
tkhs0604, line 77: Good improvement! 🙌

tkhs0604 (commented):
Thank you for your great work! It totally looks awesome!
I left 1 opinion about how to apply border, so please take a look at it 🙏

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/SessionTag.kt
Hachimori: Agreed. Fixed it!

tkhs0604 (approved):
Thank you for improving the tags!

PR #895, opened 2023-08-22
Remove bookmarkList padding

k-shinn:
## Issue
- none

## Overview (Required)
- Fixed BookmarkList padding based on Figma's design
  - padding end is none

## Links
- figma
  - https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=54902-40572&mode=dev

Before | After
:--: | :--:
 | 
adbeem-20230823005913|adbeem-20230823010544
adbeem-20230823005921|adbeem-20230823010549

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):
Looks good! Thanks for adapting layout!

PR #897, opened 2023-08-22
[iOS] Add rounded top corners to the timetable

woxtu:
## Issue
- close N/A

## Overview (Required)
- This PR adds rounded top corners to the timetable

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


ry-itto (approved):
Good catch! LGTM! Thank you!

Issue #789, opened 2023-08-20
Long session titles should be truncated with an ellipsis.


Aniokrait:
**Idea Description**
Too long session titles should be truncated with an ellipsis to fit in the grid card like in [Figma](https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=56145-49893&mode=design&t=ACPg0PqDs2wfqfEp-4). However the long title pushes schedules and speakers out of the card and it fills all the space.

**Reference images and links**
| Current behavior | Expected bahavior |
| ---- | ---- |
| <img src="https://github.com/DroidKaigi/conference-app-2023/assets/43056135/96c04a33-21d2-41d0-9560-d66f26607f2c" width=300 /> | <img width="300" src="https://github.com/DroidKaigi/conference-app-2023/assets/43056135/8dde1cf3-28ef-4c4e-8429-5fec5eeaba8a"> |

https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=56145-49893&mode=design&t=ACPg0PqDs2wfqfEp-4

takahirom:
Well, as far as I know, we should use small text. Let me consult this with the designer

takahirom:
I've checked.
We have to make it small by 12sp. And if it doesn't fit, we have to use `...`.

Noddy20:
I can take this ✋🏻
Max lines should be 4?

takahirom:
@Noddy20 
Thanks! 

> Max lines should be 4?

No, you shouldn't.

It will be a little logic. 😅 
This is just explanation code.

```kotlin
text.size = 14.sp
if(text.canFitInLayout) {
  return
}
sizeCandidate = calcFitLayout()
if(12 <= sizeCandidate) {
  text.size = sizeCandidate
  return
}
text.size = 12.sp
text.ellipsize = true
```

PR #898, opened 2023-08-22
[#789] Fixed Timetable grid item UI

Noddy20:
## Issue
- closes [#789](https://github.com/DroidKaigi/conference-app-2023/issues/789)
## Overview
[Updated] Changes made in the TimeTable grid item UI -
- Issue fixed : Long titles were truncating, Fixed with ellipsis
- Added Checks and changes to make UI look like [Figma UI](https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=54462-36316&mode=design&t=k5cPvMJl7NlCeYwU-0)
- Updated the error content (replace full message with icon only) as shown in Figma
### Background Context -
- Truncating lines fixed by the Column weights

## Links
- NA

Before | After
:--: | :--:
 | 
 |


feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableGridItem.kt
```diff
@@ -108,41 +109,67 @@ fun TimetableGridItem(
...
-                )
-            }
+        }
 
-            Spacer(
-                modifier = Modifier
-                    .weight(1f)
-                    .defaultMinSize(minHeight = 8.dp),
-            )
+        if (speakers.isEmpty()) return@ConstraintLayout
```
takahirom, line 153: Almost LGTM! I would like to use if block here. This is because we don't expect an early return when we write Compose so we could accidentally skip important component 🙏

Noddy20: Added.

takahirom:
It looks like the sessions that don't have a speaker can break. Could you check it? We have several sessions that have no speakers. 👀 
image

Noddy20:
> It looks like the sessions that don't have a speaker can break. Could you check it? We have several sessions that have no speakers. 👀 image

I'll look into it.

Noddy20:
> It looks like the sessions that don't have a speaker can break. Could you check it? We have several sessions that have no speakers. 👀 image

Fixed it, please review.

Noddy20:
> If there is a speaker. Maybe the title is going to center? image

Thanks 🧐 @takahirom, Looks like the whole content is moving center vertically when the content is less.
Pushing a fix right away.

takahirom:
Sorry for bothering you. But, the text is hidden when the item is small. What do you think?
image

takahirom:
Oh, let me rerun the CI

Noddy20:
> Sorry for bothering you. But, the text is hidden when the item is small. What do you think? image

I can take a look at that. Might need to change the approach.
@takahirom Is it possible to have a case with that small item? In a case like that what can we prioritize to show, since it will be difficult to fit all the content there.

Noddy20:
@takahirom 
- Refactored the code and made some changes according to the Figma UI
- There is one catch though, the schedule time text and speaker name title fontSizes will be adjusted according to the fontSizes of the title. Hope it works since a small title fontSize will any way look weird. This was required to make the Grid item more responsive for different heights.

takahirom:
I think we can't see the session message 👀 
image

takahirom:
Oh, I see.

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableGridItem.kt
```diff
@@ -111,14 +110,24 @@ fun TimetableGridItem(
             }
             .padding(TimetableGridItemSizes.padding),
     ) {
-        Column {
+        Column(
```
takahirom, line 113: Honestly, I was hesitant to use ConstraintLayout, as it's not compatible with other platforms. This alternative seems much better.

takahirom (approved):
Thank you for this PR. Your relentless perseverance was impressive.

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableGridItem.kt
Noddy20: Yes. I had to try multiple iterations to make this solution 😅
Best thing is that now we have the exact Figma UI 🙌🏻

PR #910, opened 2023-08-23
use Activity.enableEdgeToEdge(...)

### LINKED ISSUE #909: Can we use `Activity.enableEdgeToEdge()`?
**Idea Description**

https://github.com/android/nowinandroid/commit/2b218b97d959d79268bb0883a9e31ebcea18cb82

ked4ma:
## Issue
- close #909

## Overview (Required)
- As written in Issue, use Activity.enableEdgeToEdge(...) instead of using WindowCompat / SystemUiController

## Links
- https://github.com/android/nowinandroid/commit/2b218b97d959d79268bb0883a9e31ebcea18cb82

※Expect: No diff arround the system bar and the navigation bar
api | Before | After
:--: | :--: | :--:
34 | <br> | <br>
23 |   |

### DIFF
--- a/app-android/src/main/java/io/github/droidkaigi/confsched2023/KaigiApp.kt
+++ b/app-android/src/main/java/io/github/droidkaigi/confsched2023/KaigiApp.kt
@@ -9,13 +9,11 @@ import android.os.Build
 import android.provider.CalendarContract
 import androidx.annotation.RequiresApi
 import androidx.browser.customtabs.CustomTabsIntent
-import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Surface
 import androidx.compose.material3.windowsizeclass.WindowSizeClass
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.DisposableEffect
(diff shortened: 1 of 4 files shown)


app-android/src/main/java/io/github/droidkaigi/confsched2023/KaigiApp.kt
```diff
@@ -84,11 +82,9 @@ fun KaigiApp(
...
 ) {
     KaigiTheme {
-        val systemUiController = rememberSystemUiController()
         val useDarkIcons = !isSystemInDarkTheme()
 
-        DisposableEffect(systemUiController, useDarkIcons) {
-            systemUiController.setSystemBarsColor(Color.Transparent, useDarkIcons)
+        DisposableEffect(useDarkIcons) {
             onDispose {}
         }
```
takahirom, line 89: Maybe we can remove this? 👀

ked4ma: That's true. It do nothing :sweat_smile: 
I removed it !
https://github.com/DroidKaigi/conference-app-2023/pull/910/commits/2e408f9b579bbd9b92858e101f628a4c03c90a25

takahirom (approved):
Looks great! Thank you for introducing the new mechanism!

PR #911, opened 2023-08-23
Tweak layout spacing for the timetable screen tabs

h6ah4i:
## Overview (Required)

Tweak spacing around tabs on the timetable screen.

## Links

- [DroidKaigi 2023 App UI – Figma > Timetable/List/FV](https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=56145-49559&mode=design&t=7a5a3LtVLzDym3kX-4)

## Screenshot
Before | After
:--: | :--:
|  |  |
|  |  |

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableTab.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableTab.kt
@@ -89,7 +89,7 @@ fun TimetableTabIndicator(
     Box(
         modifier
             .zIndex(-1f)
-            .padding(horizontal = 8.dp)
+            .padding(horizontal = tabIndicatorHorizontalSpacing)
             .fillMaxSize()
             .background(
                 color = MaterialTheme.colorScheme.primary,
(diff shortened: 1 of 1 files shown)


feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableTab.kt
```diff
@@ -179,8 +181,11 @@ class TimetableTabState(
 }
 
 private val minTabHeight = 32.dp
+private val baselineTabHeight = 56.dp
 private val maxTabRowHeight = 84.dp
 private val minTabRowHeight = 56.dp
+private val tabIndicatorHorizontalSpacing = 8.dp
+private val tabRowHorizontalSpacing = (maxTabRowHeight - baselineTabHeight) / 2 - tabIndicatorHorizontalSpacing
```
takahirom, line 188: May I ask why we used `maxTabRowHeight - baselineTabHeight` here? Is it a size for the corner round?

takahirom (approved):
I'm not sure how it calculated a lot. Looks great!

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableTab.kt
h6ah4i: Let me explain :raising_hand_man:

Tabs are already vertically center aligned of the parent container element (`TabRow`), so I thought the horizontal spacing size should be the same as the vertical ones.

However, I noticed my mistake after this PR getting merged. (Sorry!) I checked the UI design spec (Figma) and it says a bit different layout constraints specified to Tabs and TabRow. They are not vertically aligned and the bottom padding amount is less than other directions.

I'm going to create a PR to match the implementation with the UI design spec later.

g112452

PR #914, opened 2023-08-23
fix Bookmark snackbar's "View" button

### LINKED ISSUE #609: Bookmark snackbar's "View" button doesn't work
**Idea Description**

We would like to see the BookmarkScreen when we click it.
image

nemo-855:
## Issue
- close #609 

## Overview (Required)
- I created TimetableItemDetailEvent to represent the snack bar display as a one-shot process.
- TimetableItemDetailViewModel managed TimetableItemDetailEvents as SharedFlow.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:
<video src="" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/77588574/58e514af-9459-4b79-b6a8-358aa24a145c" width="300" >


feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableItemDetailViewModel.kt
```diff
@@ -59,16 +63,20 @@ class TimetableItemDetailViewModel @Inject constructor(
                 isBookmarked = bookmarked,
             )
         }
+    val event: SharedFlow<TimetableItemDetailEvent> = _event
```
takahirom, line 66: Thanks! I think this is better because we don't have a leak.
But passing events will be difficult to manage. 
image
https://developer.android.com/topic/architecture/recommendations

How about "viewBookmarkRequestStateFlow = MutableStateFlow(RequestState.NotRequested)" and merge it to uiState 👀

nemo-855: @takahirom 
> How about creating stateflow "viewBookmarkListRequestStateFlow = MutableStateFlow(RequestState.NotRequested)" and merge it to uiState 👀

That's sounds great!!
I have revised that policy.This is [the commit](https://github.com/DroidKaigi/conference-app-2023/pull/914/commits/ef3f57084c1811100347be396e8dee5807738a80).

Here is the intent of this commit.

> Why `val shouldViewBookmarkList` be defined at the top level of TimetableItemDetailScreenUiState instead of in Loaded?

Since I considered the screen transition request to be a separate axis concept not related to the current screen state, I thought it was inappropriate under Loaded.

> Why did I name it ViewBookmarkListRequestState instead of RequestState?

Suppose we name it RequestState. In this case, when the number of similar one-shot requests increases in the future
It will be implemented as one of the branches of RequestState.

Since UiState can only manage one RequestState at this time, we thought it would be difficult to send two one-shot requests at the same time.

If we named it "ViewBookmarkListRequestState", it would be used exclusively for screen transition requests to BookmarkScreen, so the above problem would be less likely to occur.


Translated with www.DeepL.com/Translator (free version)

nemo-855: And it has also been verified to work properly!

<video src="https://github.com/DroidKaigi/conference-app-2023/assets/77588574/01a5977a-3beb-4a30-8c04-f01111e7034e" width="300" >

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableItemDetailScreen.kt
```diff
@@ -88,6 +92,13 @@ fun TimetableItemDetailScreen(
         userMessageStateHolder = viewModel.userMessageStateHolder,
     )
 
+    LaunchedEffect(uiState.shouldViewBookmarkList) {
+        if (uiState.shouldViewBookmarkList) {
+            navigateToBookmarkScreen()
```
takahirom, line 97: How about using onNavigateToBookmarkScreenRequested. Because we pass events to parent composable functions 👀

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableItemDetailScreen.kt
```diff
@@ -88,6 +92,13 @@ fun TimetableItemDetailScreen(
         userMessageStateHolder = viewModel.userMessageStateHolder,
     )
 
+    LaunchedEffect(uiState.shouldViewBookmarkList) {
+        if (uiState.shouldViewBookmarkList) {
```
takahirom, line 96: And could you use shouldNavigateToBookmarkList for here? 👀

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableItemDetailEvent.kt
```diff
@@ -0,0 +1,5 @@
+package io.github.droidkaigi.confsched2023.sessions
+
+sealed interface TimetableItemDetailEvent {
```
takahirom, line 3: Can we remove this? 👀

nemo-855:
@takahirom 
I have corrected the following commits

> How about using onNavigateToBookmarkScreenRequest. Because we pass events to parent composable functions 👀

> And could you use shouldNavigateToBookmarkList for the name here? 👀

[695c1ff](https://github.com/DroidKaigi/conference-app-2023/pull/914/commits/695c1ffe9ef316aa62b9b7c9adff43969d41bce4).

> Can we remove this? 👀

[ac4395e](https://github.com/DroidKaigi/conference-app-2023/pull/914/commits/ac4395e1c880e2498f1f294dd602b514a9fb9ae0)

I am very forgetful and your review was very helpful. Thank you!

takahirom (approved):
Thanks for creating the new way to handle this!

PR #916, opened 2023-08-23
Pre commit hook

Noddy20:
## Issue
- NA : No issue present, contribution to improve CI pipeline experience

## Overview
- Added pre commit git hook to run the detekt and lint check. Commit will only succeed if the pre commit checks passes.
- This helps preventing CI pipeline runs that might fail due to minor formatting or lint related issue
- This also helps first time contributors to fix all the formatting related issues before committing and pushing any change

## Links
- NA

Before | After
:--: | :--:
NA

## Movie (Optional)
Before | After
:--: | :--:
NA

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -7,6 +7,10 @@ plugins {
     alias(libs.plugins.kotlinxKover) apply false
 }
 
+apply {
+    from("${rootDir}/git-hooks.gradle.kts")
+}
+
 tasks.register("clean", Delete::class) {
     delete(rootProject.layout.buildDirectory)
 }
--- a/git-hooks.gradle.kts
+++ b/git-hooks.gradle.kts
@@ -1,0 +1,24 @@
+tasks.register("copyGitHooks", Copy::class.java) {
+    description = "Copies the git hooks from /git-hooks to the .git folder."
+    group = "git hooks"
+    from("$rootDir/hooks/pre-commit")
+    into("$rootDir/.git/hooks/")
+}
+
+tasks.register("installGitHooks", Exec::class.java) {
+    description = "Installs the pre-commit git hooks from /git-hooks."
+    group = "git hooks"
+    workingDir = rootDir
+    commandLine = listOf("chmod")
+    args("-R", "+x", ".git/hooks/")
+    dependsOn("copyGitHooks")
+    doLast {
+        logger.info("Git hook installed successfully.")
+    }
+}
+
+afterEvaluate {
+    tasks.findByName("clean")?.dependsOn(":installGitHooks")
+    tasks.findByName("assemble")?.dependsOn(":installGitHooks")
+    tasks.getByPath(":app-android:preBuild").dependsOn(":installGitHooks")
+}
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,0 +1,27 @@
+#!/bin/bash
+set -e
+echo "*******************************"
+echo   "Running git pre-commit hook"
+echo "*******************************"
+
+./gradlew lintDevDebug
+lintDevDebugStatus=$?
+echo "lintDevDebugStatus Result :"
+if [ "$lintDevDebugStatus" = 0 ]; then
+    echo "No issues were found in lintDevDebug"
+else
+    echo 1>&2 "Need to fix issues before commit."
+    exit 1
+fi
+
+
+./gradlew detekt
+detektStatus=$?
+echo "detektStatus Result :"
+if [ "$detektStatus" = 0 ]; then
+    echo "No issues were found, committing changes."
+    exit 0
+else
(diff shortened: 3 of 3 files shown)


takahirom:
Thank you. I didn't know that you could insert a hook like this with Gradle. It seems like it might also handle Gradle's cache, which is nice. This method was even discussed among the staff, but while formatting with hooks and such is certainly one way to do it, I find it a bit awkward to be kept waiting with every small commit. There might be cases where you want to format everything at once, so I think I won't merge this time. I'm really sorry, but I think it will be a good reference, so thank you.

git-hooks.gradle.kts
```diff
@@ -0,0 +1,24 @@
...
+    doLast {
+        logger.info("Git hook installed successfully.")
+    }
+}
+
+afterEvaluate {
+    tasks.findByName("clean")?.dependsOn(":installGitHooks")
+    tasks.findByName("assemble")?.dependsOn(":installGitHooks")
+    tasks.getByPath(":app-android:preBuild").dependsOn(":installGitHooks")
+}
```
jmatsu, line 24: You can change the git hooks directory from `.git/hooks` to any directory you want. It's better to use the custom hook directory rather than copying hook scripts frequently.

Please add these lines to /.gitignore.

```
# Ignore files under /hooks except pre-defined hooks
/hooks/*
!/hooks/pre-commit
```

and then, create a task in the */build.gradle.kts*.

```suggestion
tasks.register("installGitHooks", Exec::class.java) {
    description = "Set up local git config to use /hooks for git-hooks"
    group = "git hooks"
    workingDir = layout.projectDirectory.getAsFile()
    commandLine = listOf("git")
    args("config", "--local", "core.hooksPath", "hooks")
}
```

Lastly, please make reference to `./gradlew installGitHooks` in [README.md#Requirements](https://github.com/DroidKaigi/conference-app-2023#requirements).

hooks/pre-commit
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+set -e
```
jmatsu, line 2: `exec 1>&2` is a better practice that redirects stdout to stderr in hooks because some of users may wanna use stdout of `git commit`.

```suggestion
set -e

exec 1>&2
```

hooks/pre-commit
```diff
@@ -0,0 +1,27 @@
...
+
+./gradlew lintDevDebug
+lintDevDebugStatus=$?
+echo "lintDevDebugStatus Result :"
+if [ "$lintDevDebugStatus" = 0 ]; then
+    echo "No issues were found in lintDevDebug"
+else
+    echo 1>&2 "Need to fix issues before commit."
+    exit 1
+fi
```
jmatsu, line 15: `-e` option means this script(shell) will abort right after any non-zero code is returned so `$?` without if-statement won't work. 

```suggestion
if ./gradlew lintDevDebug then
    echo "No issues were found in lintDevDebug"
else
    echo "Android Lint exited with $?. Need to fix issues before commit."
    exit 1
fi
```

hooks/pre-commit
```diff
@@ -0,0 +1,27 @@
...
+./gradlew detekt
+detektStatus=$?
+echo "detektStatus Result :"
+if [ "$detektStatus" = 0 ]; then
+    echo "No issues were found, committing changes."
+    exit 0
+else
+    echo 1>&2 "Need to fix issues before commit."
+    exit 1
+fi
```
jmatsu, line 27:
```suggestion
if ./gradlew detekt then
    echo "No issues were found, committing changes."
else
    echo "Detekt exited with $?. Need to fix issues before commit."
    exit 1
fi
```

build.gradle.kts
```diff
@@ -7,6 +7,10 @@ plugins {
     alias(libs.plugins.kotlinxKover) apply false
 }
 
+apply {
+    from("${rootDir}/git-hooks.gradle.kts")
+}
+
```
jmatsu, line 13: Please define the install task directly.

```suggestion
```

jmatsu (changes requested):
Although we still continue to discuss this topic as described in https://github.com/DroidKaigi/conference-app-2023/pull/916#issuecomment-1689953649, I personally love Git hooks solution so let me submit a couple of reviews in advance.

build.gradle.kts
Noddy20: Any reason for defining it directly?
I kept a separate file for clean code (Separation of Concerns) and better visibility.

takahirom: I think you can refer to this.
https://developer.squareup.com/blog/herding-elephants/#model-your-build

takahirom:
We might use these scripts on other projects but for now, we don't use this. Let me close. Thank you for your suggestion and work.

PR #917, opened 2023-08-23
Fix Find NORMAL Session

### LINKED ISSUE #915: Find a normal session for TimetableItemDetailScreenTest
**Idea Description**
In this PR, we are having welcome session in screenshots. However, I would like to see the normal session screenshots.
So could we do `SessionsAllResponse.fake().sessions.first{ it.xxxxType == Normal }.id  ` or something to find the normal session?
https://github.com/DroidKaigi/conference-app-2023/pull/822
image

https://github.com/DroidKaigi/conference-app-2023/blob/39f785294d84e6e44017f6160e2e4438e80ab923/core/testing/src/main/java/io/github/droidkaigi/confsched2023/testing/robot/TimetableItemDetailScreenRobot.kt#L93

shinya-takano:
## Issue
- close #915

## Overview (Required)
- Fix Find NORMAL Session

## Links
none.

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/core/testing/src/main/java/io/github/droidkaigi/confsched2023/testing/robot/TimetableItemDetailScreenRobot.kt
+++ b/core/testing/src/main/java/io/github/droidkaigi/confsched2023/testing/robot/TimetableItemDetailScreenRobot.kt
@@ -105,6 +105,6 @@ class TimetableItemDetailScreenRobot @Inject constructor(
     }
 
     companion object {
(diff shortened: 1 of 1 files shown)


takahirom:
Let me merge this.
We have bloken main 🙏 
Thanks for your prompt fix

PR #822, opened 2023-08-21
Implementation fake response of "WELCOME_TALK"

sanao1006:
## Issue
- close #667 

## Overview (Required)
- Added "Welcome Talk" session to FakeSessionsApiClient to facilitate various testing scenarios

## Concern
- Are the "startsAt" and "endsAt" times not appropriate?(https://github.com/DroidKaigi/conference-app-2023/issues/667#issuecomment-1685985567)
    > But can you use this time to see it in the screenshot tests?
startsAt":"2023-09-14T10:30:00+09:00","endsAt":"2023-09-14T11:00:00+09:00" (This is the start time of DroidKaigi)
- Should I use enum for sessionType?

PR #919, opened 2023-08-23
Revert "Implementation fake response of "WELCOME_TALK""

takahirom:
Reverts DroidKaigi/conference-app-2023#822


takahirom:
We were able to fix the test 🙏

PR #923, opened 2023-08-23
Configure workflows to use proper run-names and for DRY

jmatsu:
## Issue
- close N/a

## Overview (Required)

- Specify `run-name` for each workflow unless defined
- Aggregate a couple of setup actions for iOS build into the single action

## Links
- N/a


takahirom:
Thanks, it looks good, but I'm sorry, I maybe don't understand the merit. 🙇  It seems that the information we need is already available. Additionally, I would like to find an action by title. 👀 
image

PR #930, opened 2023-08-23
Set JST time zone for robot tests

### LINKED ISSUE #906: Set JST Timezone for tests
**Idea Description**
We might have it to RobotTest

This should be 10:30 am

image

chibatching:
## Issue
- close #906

## Overview (Required)
- Set JST time zone for all robot test as test rule

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/core/testing/src/main/java/io/github/droidkaigi/confsched2023/testing/TimeZoneTestRule.kt
+++ b/core/testing/src/main/java/io/github/droidkaigi/confsched2023/testing/TimeZoneTestRule.kt
@@ -4,9 +4,15 @@ import org.junit.rules.TestWatcher
 import org.junit.runner.Description
 import java.util.TimeZone
 
-class TimeZoneTestRule : TestWatcher() {
+class TimeZoneTestRule() : TestWatcher() {
+    private lateinit var evacuatedTimeZone: TimeZone
+
     override fun starting(description: Description) {
-        super.starting(description)
+        evacuatedTimeZone = TimeZone.getDefault()
         TimeZone.setDefault(TimeZone.getTimeZone("Asia/Tokyo"))
+    }
+    
+    override fun finished(description: Description) {
+        TimeZone.setDefault(evacuatedTimeZone)
     }
 }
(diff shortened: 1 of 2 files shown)


core/testing/src/main/java/io/github/droidkaigi/confsched2023/testing/TimeZoneTestRule.kt
```diff
@@ -0,0 +1,12 @@
...
+import org.junit.rules.TestWatcher
+import org.junit.runner.Description
+import java.util.TimeZone
+
+class TimeZoneTestRule : TestWatcher() {
+    override fun starting(description: Description) {
+        super.starting(description)
+        TimeZone.setDefault(TimeZone.getTimeZone("Asia/Tokyo"))
+    }
+}
```
jmatsu, line 12: Global configuration may cause an problem so it should be reverted on finished anyway. And also, `super` don't need to be called in TestWatcher by the way.

```suggestion
class TimeZoneTestRule() : TestWatcher() {
    private lateinit var evacuatedTimeZone: TimeZone

    override fun starting(description: Description) {
        evacuatedTimeZone = TimeZone.getDefault()
        TimeZone.setDefault(TimeZone.getTimeZone("Asia/Tokyo"))
    }
    
    override fun finished(description: Description) {
        TimeZone.setDefault(evacuatedTimeZone)
    }
}
```

jmatsu (changes requested):
The approach looks right. Could you please check my review?

core/testing/src/main/java/io/github/droidkaigi/confsched2023/testing/TimeZoneTestRule.kt
takahirom: Looks good! Thanks!

takahirom (approved):
I would do the same implementation! Thanks!

momomomo111 (approved):
Thanks! LGTM!

PR #932, opened 2023-08-24
Adding tests to BookmarkScreenTest

### LINKED ISSUE #894: Adding tests to BookmarkScreenTest
**Idea Description**
The title says everything.
We might have to use UserDataStore to adjust the bookmark.

https://github.com/DroidKaigi/conference-app-2023/pull/868

hiesiea:
## Issue
- close #894 

## Overview (Required)
- Adding tests to BookmarkScreenTest

## Links
- none

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


core/testing/src/main/java/io/github/droidkaigi/confsched2023/testing/robot/BookmarkScreenRobot.kt
```diff
@@ -47,6 +60,36 @@ class BookmarkScreenRobot @Inject constructor(
...
+
+    fun clickBookmarkFilterChipDay3() {
+        composeTestRule
+            .onNode(hasTestTag(BookmarkFilterChipDay3TestTag))
+            .performClick()
+    }
+
+    fun toggleFavorite() = runTest(testDispatcher) {
+        SessionsAllResponse.fake().toTimetable().bookmarks.forEach {
+            userDataStore.toggleFavorite(it)
```
hiesiea, line 89: Sorry, I have a question.
I tried to write a value using UserDataStore#toggleFavorite, but when I debugged it, it didn't seem to be written.
Could you please tell me what I should do?

takahirom: I think we can add suspend here. Could you check it?
image

hiesiea: Thank you very much. However, when I ran the test, I got an error.

```
Detected use of different schedulers. If you need to use several test coroutine dispatchers, create one TestCoroutineScheduler and pass it to each of them.
```

The data used was incorrect and has been fixed in e83a729.

takahirom: May I ask if you use `suspend fun toggleFavorites` and `block: suspend BookmarkScreenRobot.() -> Unit` both ?

hiesiea: Yes, I used the following code:

```kotlin
class BookmarkScreenRobot @Inject constructor(
    private val testDispatcher: TestDispatcher,
) {
    ...
    operator fun invoke(
        block: BookmarkScreenRobot.() -> Unit,
    ) {
        runTestWithLogging(timeout = 30.seconds) {
            this@BookmarkScreenRobot.composeTestRule = robotTestRule.composeTestRule
            toggleFavorites()
            block()
        }
    }
    ...
    suspend fun toggleFavorites() {
        SessionsAllResponse.fake().toTimetable().timetableItems.forEach {
            userDataStore.toggleFavorite(it.id)
        }
    }
}
```

takahirom (approved):
Thanks for trying adjusting and your cooperation!

PR #937, opened 2023-08-24
Changed the priority of repository to search for plugins

NUmeroAndDev:
## Issue
- None

## Overview (Required)
Changed the priority of repository to search for plugins.  
It may not be much of a problem, but gradlePluginPortal should have a lower priority to solve, so I changed it.

## Links
- https://developer.android.com/build/optimize-your-build#gradle_plugin_portal

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -1,9 +1,9 @@
 pluginManagement {
     includeBuild("build-logic")
     repositories {
-        gradlePluginPortal()
         google()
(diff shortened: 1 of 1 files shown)


jmatsu (approved):
Makes sense. LGTM! 🎉

takahirom:
Thanks!

PR #940, opened 2023-08-24
Use `TYPESAFE_PROJECT_ACCESSORS` for `core/testing` module

fumiya-kume:
## Issue

- N/A

## Overview (Required)
- Use `TYPESAFE_PROJECT_ACCESSORS` for `core/testing` module


takahirom (approved):
It's definitely better

PR #943, opened 2023-08-24
Add codeowner for iOS related PRs

tkhs0604:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- SSIA

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):
Looks good.👍 I'm not familiar with this setting though😂

PR #945, opened 2023-08-24
Remove unnecessary code that just return boolean directly.

fumiya-kume:
## Issue
- N/A

## Overview (Required)
- Remove unnecessary code that just return boolean directly.
- I think no reason to keep current code?


takahirom (approved):
That makes sense!

Issue #761, opened 2023-08-19
Create a single coverage report by KotlinX Cover


takahirom:
**Idea Description**
We can specify dependency for kover
https://kotlin.github.io/kotlinx-kover/gradle-plugin/

```
dependencies {
    kover(project(":core"))
}
```

> for example, in the :app project, the :app:koverHtmlReport task call will generate one HTML report containing the :app and :core classes.

And we can see some examples to do it.

This is a script in coroutines. This script is used in root build.gradle.

https://github.com/Kotlin/kotlinx.coroutines/blob/1ccc96890d38be52d1bdcd31af5befb54804aa05/buildSrc/src/main/kotlin/kover-conventions.gradle.kts#L27

We would like to create a plugin in https://github.com/DroidKaigi/conference-app-2023/tree/main/build-logic/src/main/kotlin/io/github/droidkaigi/confsched2023/primitive like this

**Reference images and links**

takahirom:
Thanks! We need this!

PR #948, opened 2023-08-24
Aggregate kover reports into single report

nashcft:
## Issue
- #761 (part 1/2)
  - next one is to extract kover configuration to build-logic

## Overview (Required)
- Aggregate kover coverage reports into one report for `:app-android` so that we can generate the coverage report for all classes in all modules with kover report task of `:app-android`
- Add filter configuration to exclude coverage reports for generated code from reports

## Links
- Kover doc: https://kotlin.github.io/kotlinx-kover/gradle-plugin/

## Known Issue

Currently coverage reports for generated code are not completely excluded although matched filter configurations exist.

N/A
## Movie (Optional)
N/A


takahirom (approved):
Looks great! Thank you for visualizing our effort!

PR #949, opened 2023-08-24
Remove duplicate dependency at feature module for core module

fumiya-kume:
## Issue
- N/A

## Overview (Required)
- Remove duplicate dependency at feature module for core module
- Currently, our architecture want to that feature module refer the some core module.
- However, we can try to reduce the duplicate code that dependencies setting.
   - For this purpose, I tried to move the code into the feature module plugin.

## Links
-


takahirom:
We may have a module that doesn't have to have them so let me close 🙏  Thanks for a lot of contribution!

PR #952, opened 2023-08-24
[iOS] Fix the layout of the timetable item

woxtu:
## Issue
- close N/A

## Overview (Required)
- This PR fixes the layout of the timetable item

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


tkhs0604 (approved):
I remember removing this line myself 😭, thank you for fixing the layout!

PR #959, opened 2023-08-25
Consider WindowInset

### LINKED ISSUE #956: WindowInsets is ignored
**Describe the bug**
WindowInsets are ignored and hidden behind NavigationBar

**To Reproduce**
Open Screen
- Timetable
  - List
  - Gid
- Floor Map
- Stamps(maybe)
- About
- Contributors

**Expected behavior**
The UI is displayed without being hidden by the system bar.

**Screenshots**
bottom is filled.

right is filled.

matsudamper:
## Issue
- close #956

## Overview (Required)
- Made WindowInsets propagate from parent.

## Links
- 

<table>
<tr>
<td>before</td>
<td>after</td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Movie (Optional)
Before | After
:--: | :--:


matsudamper:
The `BookmarkTopArea` has a fixed height and the layout is broken by the size of the text. It is better to use `LargeTopAppBar`, I thought it would be better to change it to the same implementation as `TimetableItemDetailScreenTopAppBar`.

matsudamper:
on purpose
- Bookmark
  - The dimensions of the SystemBar are provided by WindowInsets.

confirm
- FloorMap
  - shadow
  - mobile layout 
 - Timeline

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableScreen.kt
```diff
@@ -207,6 +222,11 @@ private fun TimetableScreen(
                 uiState = uiState.contentUiState,
                 timetableScreenScrollState = state,
                 onFavoriteClick = onBookmarkClick,
+                windowInsetsPadding = PaddingValues(
+                    bottom = innerPadding.calculateBottomPadding(),
```
takahirom, line 226: May I ask if `innerPadding.calculateBottomPadding()` contains OS's navigatiaon bar and bottom navigation size? 🙋

takahirom: I don't understand yet 🙇

matsudamper: What is passed from Scaffold is a recognition that the sizes of Systembar and BottomBar are combined.

matsudamper: This naming is out of sync with the others and should be changed.
windowInsetsPadding → contentPadding

takahirom: So the inner layout has the bigger padding, right?

Would it be good to adapt using rules like these?

## The Policy
Insets are not consumed; instead, they are added. As a result, the inner padding will be larger.

## The Rule
Insets should be passed as parameters to screen-level Composable functions.
Composable functions for screen containers, such as MainScreen, should add insets to child screens and pass them along.

matsudamper: I think it's a good rule to implement EdgeToEdge, Tablet and Mobile layouts.

matsudamper:
This should make Gradient look the same as before, but diffs seem to appear.
I didn't understand why the diff occurs.
https://github.com/DroidKaigi/conference-app-2023/pull/959/commits/1a7fa8f923a6213937138a7476aa60ea96e79baa

diff


If there are no design issues, I would like to revert the following to reduce complexity. Please let me know what you think.
[1a7fa8f](https://github.com/DroidKaigi/conference-app-2023/pull/959/commits/1a7fa8f923a6213937138a7476aa60ea96e79baa)

matsudamper:
I'm not sure why the following differences occur

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableScreen.kt
takahirom: Great!

takahirom:
Please fix the conflict 🙏

matsudamper:
fixed conflict.

takahirom (approved):
Thanks for how to create our app!

PR #962, opened 2023-08-25
[iOS] Make the description on the session detail selectable

### LINKED ISSUE #885: [iOS] Enable to copy main description on timetable detail screen
## Summary
Now we cannot select and copy session's main description, so make it selectable.

woxtu:
## Issue
- close #885

## Overview (Required)
- This PR makes the description on the session detail selectable to copy it

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


ry-itto (approved):
LGTM! Thanks!

PR #970, opened 2023-08-26
Support large font for session tag and timetable tab

Aniokrait:
## Issue
- sorry, I haven't opened any issue. Please tell me if I should open it before PR and I'll create the one.
- I've found the problem that the session tags and the timetable tabs don't support large font size.
  - Please refer to the screenshot I attached bottom to see exactly what happend.
### how to reproduce: 
- Device: Pixel6 emulator
- Change the font size to the largest in the device settings.
- Open the app.

## Overview (Required)
- Stopped specifying a fixed height and instead set a minimum height.

### large font size
Before | After
:--: | :--:
 | 
### normal font size
There is no difference.
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/SessionTag.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/SessionTag.kt
@@ -10,6 +10,7 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.text.font.FontWeight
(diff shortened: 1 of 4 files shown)


takahirom:
Could you add a screenshot test so that you can confirm the large font?

https://github.com/DroidKaigi/conference-app-2023/blob/3395eaf2e5212ca236114a9942a700dfc6bb483d/feature/sessions/src/test/java/io/github/droidkaigi/confsched2023/sessions/TimetableItemDetailScreenTest.kt#L120

You can add it here

https://github.com/DroidKaigi/conference-app-2023/blob/250b7bb2d4b4b8aafaeffe45bc0dc3934d225e91/feature/sessions/src/test/java/io/github/droidkaigi/confsched2023/sessions/TimetableScreenTest.kt#L41

Aniokrait:
Sure! I'll add the tests.

Aniokrait:
Thanks to your precice instructions, I could write tests easily!
By the way, the original test case names looked like not matching the naming convention. It might be none of my business, but I also corrected those names.

takahirom (approved):
Looks great!

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/SessionTag.kt
```diff
@@ -29,7 +30,7 @@ fun SessionTag(
 ) {
     Box(
         modifier = modifier
-            .height(24.dp)
+            .defaultMinSize(minHeight = 24.dp)
```
takahirom, line 33: 👍

PR #976, opened 2023-08-26
:recycle: Adding animation when stamp is pressed

### LINKED ISSUE #900: Adding animation when stamp is pressed
**Idea Description**
Please add an animation when the stamp is pressed.
We have both json and lottie files. Use whichever you prefer. Lottie files are usually smaller.
(For the timing when the stamp is pressed, we currently haven't implemented it. I'd appreciate it if you could set it up so that when a faint-colored animal icon is clicked, the animation runs.)
https://github.com/DroidKaigi/conference-app-2023/tree/main/feature/stamps/src/main/res/raw

- [Figma](https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=55666-94384&mode=dev)

Corvus400:
## Issue
- close #900

## Overview (Required)
- Tapping a stamp now displays an animation.

## Movie
Light | Dark
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/5bd14e5a-ab57-4732-a637-f724a1a9dcc1" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/b106eebc-5a5d-4827-875f-2a6c948e6333" width="300" >

### DIFF
--- a/feature/stamps/src/main/java/io/github/droidkaigi/confsched2023/stamps/section/StampList.kt
+++ b/feature/stamps/src/main/java/io/github/droidkaigi/confsched2023/stamps/section/StampList.kt
@@ -45,7 +45,10 @@ fun StampList(
         } else {
             Color.Transparent
         },
-        animationSpec = tween(400, easing = EaseInOut),
+        animationSpec = tween(
+            delayMillis = 2,
+            durationMillis = 400, easing = EaseInOut
+        ),
         label = "",
     )
     Box(
--- a/feature/stamps/src/main/java/io/github/droidkaigi/confsched2023/stamps/section/StampList.kt
+++ b/feature/stamps/src/main/java/io/github/droidkaigi/confsched2023/stamps/section/StampList.kt
@@ -1,5 +1,8 @@
 package io.github.droidkaigi.confsched2023.stamps.section
 
+import androidx.compose.animation.animateColorAsState
+import androidx.compose.animation.core.EaseInOut
+import androidx.compose.animation.core.tween
 import androidx.compose.foundation.background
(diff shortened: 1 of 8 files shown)


feature/stamps/src/main/java/io/github/droidkaigi/confsched2023/stamps/section/StampList.kt
```diff
@@ -21,39 +33,75 @@ private const val DoubleItemSpanCount = 2 / 2
...
-            horizontal = 16.dp,
-            vertical = 20.dp,
+    val bgColor: Color by animateColorAsState(
+        if (stampLottieRawId != null) {
+            Color(0xFF37383D)
+        } else {
+            Color.Transparent
+        },
+        animationSpec = tween(
+            delayMillis = 2,
```
Corvus400, line 49: In Figma, there is an instruction to specify 1.5 ms.
However, delayMillis can only specify an Int value, so it is set to 2.
I am not sure if this is correct, so please point out if I am wrong.🙇‍♂️

mhidaka: Thank you for your report and we'll talk about this issue with design's side.
But I think your solution makes sense 😉

momomomo111: We discussed the animation you implemented with our designer.
It's a great implementation and there are no design concerns 👍

momomomo111 (approved):
LGTM!

PR #980, opened 2023-08-27
The TimetableItemDetail Snackbar doesn't disappear after some seconds

### LINKED ISSUE #969: The TimetableItemDetail Snackbar doesn't disappear after some seconds
**Idea Description**
Sorry if this is unnecessary.

`SpeakerSection` is hidden by snackbar when `TimetableItemDetail` Bookmark is pressed, I thought it would be better to display space for snackbar when snackbar is show so that the `SpeakerSection` is not hidden.

I think I can detect `SnackbarHostState` and create space by doing a bucket relay or using `CompositionLocalProvider`.

kosenda:
## Issue
- close #969 

#### Added ability to set duration of snackbar with action labels

- The default duration of Material3's showSnackbar was 4 seconds if there was an action label and infinite if not.
- When the duration is set before the function `SnackbarMessageEffect` that calls Material3's `showSnackbar` is called, the value is passed to Material3's `showSnackbar`'s duration, otherwise it is left at its default value.
- Only `TimetableItemDetailViewModel`'s `onBookmarkClick` sets the snack bar duration inside the app, so it should not affect other snackbars

## Movie
Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/60963155/eda9c334-ee4c-4622-ae94-aea1af2f1cf6" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/60963155/6e6a9e97-48a9-4ec0-bd0f-d256d0304b03" width="300" >

### DIFF
--- a/core/ui/src/androidMain/kotlin/io/github/droidkaigi/confsched2023/ui/UserMessage.kt
+++ b/core/ui/src/androidMain/kotlin/io/github/droidkaigi/confsched2023/ui/UserMessage.kt
@@ -1,5 +1,6 @@
 package io.github.droidkaigi.confsched2023.ui
 
+import androidx.compose.material3.SnackbarDuration
 import androidx.compose.material3.SnackbarHostState
 import androidx.compose.material3.SnackbarResult
 import androidx.compose.material3.SnackbarResult.ActionPerformed
@@ -25,10 +26,18 @@ fun SnackbarMessageEffect(
 ) {
     userMessageStateHolder.messageUiState.userMessages.firstOrNull()?.let { userMessage ->
         LaunchedEffect(userMessage) {
-            val snackbarResult: SnackbarResult = snackbarHostState.showSnackbar(
-                message = userMessage.message,
-                actionLabel = userMessage.actionLabel,
-            )
+            val snackbarResult: SnackbarResult = if (userMessage.duration == null) {
+                snackbarHostState.showSnackbar(
+                    message = userMessage.message,
+                    actionLabel = userMessage.actionLabel,
+                )
+            } else {
+                snackbarHostState.showSnackbar(
+                    message = userMessage.message,
+                    actionLabel = userMessage.actionLabel,
+                    duration = userMessage.duration,
+                )
+            }
             userMessageStateHolder.messageShown(
                 messageId = userMessage.id,
                 userMessageResult = when (snackbarResult) {
@@ -43,6 +52,7 @@ fun SnackbarMessageEffect(
 data class UserMessage(
     val message: String,
     val actionLabel: String? = null,
+    val duration: SnackbarDuration? = null,
     val id: Long = UUID.randomUUID().mostSignificantBits,
     val userMessageResult: UserMessageResult? = null,
 )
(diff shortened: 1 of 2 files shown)


core/ui/src/androidMain/kotlin/io/github/droidkaigi/confsched2023/ui/UserMessage.kt
```diff
@@ -99,7 +113,11 @@ class UserMessageStateHolderImpl : UserMessageStateHolder {
 interface UserMessageStateHolder {
     val messageUiState: MessageUiState
     fun messageShown(messageId: Long, userMessageResult: UserMessageResult)
-    suspend fun showMessage(message: String, actionLabel: String? = null): UserMessageResult
+    suspend fun showMessage(
+        message: String,
+        actionLabel: String? = null,
+        duration: SnackbarDuration? = null,
```
takahirom, line 119: I think the default parameter for the snackbar can be Short since the official implementation is doing so.
`duration: SnackbarDuration = SnackbarDuration.Short`

core/ui/src/androidMain/kotlin/io/github/droidkaigi/confsched2023/ui/UserMessage.kt
```diff
@@ -25,10 +26,18 @@ fun SnackbarMessageEffect(
 ) {
     userMessageStateHolder.messageUiState.userMessages.firstOrNull()?.let { userMessage ->
         LaunchedEffect(userMessage) {
-            val snackbarResult: SnackbarResult = snackbarHostState.showSnackbar(
-                message = userMessage.message,
-                actionLabel = userMessage.actionLabel,
-            )
+            val snackbarResult: SnackbarResult = if (userMessage.duration == null) {
+                snackbarHostState.showSnackbar(
```
takahirom, line 30: Could you elaborate on why the snackbar doesn't disappear?  I thought the default value of snackbarHostState.showSnackbar's duration is Short. So I thought it doesn't have change 👀

kosenda: The reason why the snack bar did not disappear is because the duration of snackbar is not set anywhere in the application, and default value for duration of `showSnackbar` in `Material3` is infinite if there is an `actionLabel`, and 4 seconds otherwise, and this time snackbar had an `actionLabel` set.

image

core/ui/src/androidMain/kotlin/io/github/droidkaigi/confsched2023/ui/UserMessage.kt
kosenda: As you can see below, the `duration` of `showSnackbar` in `Material3` seems to change its default value depending on whether there is an actionLabel or not, so I thought it would be better not to define it, but it would be better to set `SnackbarDuration.Short` ?

core/ui/src/androidMain/kotlin/io/github/droidkaigi/confsched2023/ui/UserMessage.kt
takahirom: That makes sense!

takahirom (approved):
Thanks for your opinion!

PR #982, opened 2023-08-27
Bookmarking with TimetableItemDetail automatically shrinks the DescriptionSection

### LINKED ISSUE #967: Bookmarking with `TimetableItemDetail` automatically shrinks the `DescriptionSection`
**Describe the bug**
When `ReadMoreOutlinedButton` is pressed in `TimetableItemDetail` and the DescriptionSection is expanded, pressing the bottom bookmark will automatically shrink `DescriptionSection`.

**To Reproduce**
1. Go to `TimetableItemDetail`
2. Click on `ReadMoreOutlinedButton`
3. Click on BookmarkIcon

**Expected behavior**
`DescriptionSection` does not shrink when Bookmark is pressed

**Screenshots**
Emulator(Pixel4 API28)

https://github.com/DroidKaigi/conference-app-2023/assets/60963155/8c03e1ad-0c96-4a54-b6e7-1dcc187cc523

iiinaiii:
## Issue
- close #967 

### Cause of bug
- When creating a TimetableItemDetail, uiState was specified as the target of AnimatedContent, and TimetableItemDetail was being recomposed whenever uiState's isBookmarked changed.
### Fixes
- Specify { uiState is Loaded } in the contentKey lambda of AnimatedContent so that it is recomposed only when the uiState changes to Loaded.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/16633277/0d058ef3-02bb-472b-9fe2-d101bca8bf9a" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/16633277/4190836f-052f-484e-a961-9e326e4159c0" width="300" >

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableItemDetailScreen.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableItemDetailScreen.kt
@@ -32,6 +32,7 @@ import io.github.droidkaigi.confsched2023.model.TimetableItem
 import io.github.droidkaigi.confsched2023.model.TimetableItem.Session
 import io.github.droidkaigi.confsched2023.model.fake
 import io.github.droidkaigi.confsched2023.sessions.TimetableItemDetailScreenUiState.Loaded
+import io.github.droidkaigi.confsched2023.sessions.TimetableItemDetailScreenUiState.Loading
 import io.github.droidkaigi.confsched2023.sessions.component.TimetableItemDetailBottomAppBar
 import io.github.droidkaigi.confsched2023.sessions.component.TimetableItemDetailScreenTopAppBar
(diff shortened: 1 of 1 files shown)


takahirom:
I didn't know this behavior 😮

iiinaiii:
I didn't know as well 😅
But after looking at the AnimatedContent documentation, I thought it sounded like a cause.

> If targetState is expected to mutate frequently and not all mutations should be treated as target state change, consider defining a mapping between targetState and a key in contentKey. As a result, transitions will be triggered when the resulting key changes. In other words, there will be no animation when switching between targetStates that share the same same key. By default, the key will be the same as the targetState object.

momomomo111 (approved):
LGTM!

PR #992, opened 2023-08-28
[iOS] Fix the size of the navigation bar items on the timetable

woxtu:
## Issue
- close N/A

## Overview (Required)
- This PR fixes the size of the navigation bar items on the timetable

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


ry-itto (approved):
LGTM! Thanks 🙏🏼

PR #993, opened 2023-08-28
Revive TimetableShimmerListItemPreview (Disable Shimmer only during Robolectric execution)

### LINKED ISSUE #778: Shimmer effect CI unit test fails (does not finish).
**Describe the bug**
CI unit test fails due to `TimetableShimmerListItemPreview()` in TimetableShimmerListItem.kt. The test does not finish, times out and fails. Perhaps the test does not finish because it keeps waiting for the Shimmer Effect to finish animating.  

**Hint that may possibly lead to a solution**
  - takahirom found the following log (https://twitter.com/new_runnable/status/1692887907420479687)
  > "io.github.droidkaigi.confsched2023.sessions.component_null_Theme_TimetableShimmerListItemPreview-LightMode-8 _8_null"

**To Reproduce**
1.Uncomment TimetableShimmerListItemPreview() in TimetableShimmerListItem.kt
https://github.com/DroidKaigi/conference-app-2023/blob/2a774f0e97e4500603b9eda1148cfa3bfa57e8e5/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableShimmerListItem.kt#L70-L80
2.Running tests on Github Actions, and then they fail (tests may run fine on local).

**Expected behavior**
Simmer effect CI test passes without problems

**References**
- [Log of failed tests](https://github.com/DroidKaigi/conference-app-2023/actions/runs/5908387645/job/16027760011)
- #500
- #711 
- [ShimmerEffect library in use](https://github.com/valentinilk/compose-shimmer)

yamacraft:
## Issue
- close #778 

## Overview (Required)
- I've configured Shimmer not to operate during Robolectric execution and enabled the display of TimetableShimmerListItemPreview again.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableShimmerListItem.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableShimmerListItem.kt
@@ -11,6 +11,7 @@ import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material3.Divider
+import androidx.compose.material3.Surface
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
@@ -21,12 +22,19 @@ import androidx.compose.ui.unit.dp
 import com.valentinilk.shimmer.ShimmerBounds
 import com.valentinilk.shimmer.rememberShimmer
 import com.valentinilk.shimmer.shimmer
+import io.github.droidkaigi.confsched2023.designsystem.preview.MultiLanguagePreviews
+import io.github.droidkaigi.confsched2023.designsystem.preview.MultiThemePreviews
+import io.github.droidkaigi.confsched2023.designsystem.theme.KaigiTheme
 
 const val TimetableShimmerListItemTestTag = "TimetableShimmerListItemList"
 
 @Composable
 fun TimetableShimmerListItem(modifier: Modifier = Modifier) {
-    val shimmerInstance = rememberShimmer(shimmerBounds = ShimmerBounds.View)
+    val shimmerInstance = if (isRobolectric().not()) {
+        rememberShimmer(shimmerBounds = ShimmerBounds.View)
+    } else {
+        null
+    }
     Column(
         modifier = modifier.testTag(TimetableShimmerListItemTestTag),
     ) {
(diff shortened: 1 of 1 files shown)


feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableShimmerListItem.kt
```diff
@@ -67,14 +87,23 @@ fun TimetableShimmerListItem(modifier: Modifier = Modifier) {
...
-// }
+private fun isRobolectric(): Boolean {
+    @Suppress("SwallowedException")
+    return try {
+        Class.forName("org.robolectric.Robolectric")
+        true
+    } catch (e: ClassNotFoundException) {
+        false
+    }
+}
```
yamacraft, line 98: To be honest, this isn't an ideal way to determine it.
I would appreciate it if Robolectric itself provided an API to determine if it's currently being executed... 😔

takahirom: I really use this implementation(Class.forName for checking tests). But could you check `isTest()` method?

yamacraft: Oops! I did not know that function existed. Thanks!

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableShimmerListItem.kt
```diff
@@ -21,12 +22,20 @@ import androidx.compose.ui.unit.dp
...
+import io.github.droidkaigi.confsched2023.designsystem.preview.MultiThemePreviews
+import io.github.droidkaigi.confsched2023.designsystem.theme.KaigiTheme
+import io.github.droidkaigi.confsched2023.ui.isTest
 
 const val TimetableShimmerListItemTestTag = "TimetableShimmerListItemList"
 
 @Composable
 fun TimetableShimmerListItem(modifier: Modifier = Modifier) {
-    val shimmerInstance = rememberShimmer(shimmerBounds = ShimmerBounds.View)
+    val shimmerInstance = if (isTest().not()) {
```
takahirom, line 34: Could you add a comment for this? 👀

yamacraft: done!

takahirom (approved):
Looks great! Thanks!

PR #1003, opened 2023-08-29
(Proposal) Added animation to the display when the Read More button is tapped.

Corvus400:
## Issue
- None.

## Overview (Required)
- The title is telling.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/0c49ff45-b395-4d92-a399-37c0eac2a434" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/b1c70114-2cbe-46d7-a22f-6b7fa3d8743f" width="300" >


takahirom:
Sorry for saying this, but I think the old one is better because we can see if the text has the same content by not showing the animation. 🙇 I think it is similar to the shared element transition.
(And looks like the next item is going up a little while animation is running 👀 )

PR #1004, opened 2023-08-29
:wrench: Prevents clicks during animation playback

Corvus400:
## Issue
- None.

## Overview (Required)
- The title says it all.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/ac7e2e65-c63c-4960-88a2-0c21037c682c" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/07c4ff6b-d1de-4f72-8130-65d0f3e806ec" width="300" >


momomomo111 (approved):
Thank you for your contribution!
I've left a comment on one point.

takahirom (approved):
Thanks!

momomomo111:
It's an edge case, but I think this solution would work for consecutive taps of the tag🙆‍♂️

momomomo111:
> I don't think we run animation on click when the app releases 👀 So I'll leave it to @momomomo111

It's an edge case, but I think this solution would work for consecutive taps of the tag🙆‍♂️

PR #1006, opened 2023-08-29
:memo: PNG converted to WebP.

Corvus400:
## Issue
- None.

## Overview (Required)
- Only the Android side was changed because we are not sure if the iOS side can use WebP.


takahirom (approved):
Looks great! Now we have small images!

PR #1011, opened 2023-08-29
[feature/#719] Add Benchmark Module and baseline profile generator

### LINKED ISSUE #719: Add benchmark module and setup Baseline Profile generator
**Idea Description**

If we add the benchmark module, we can introduce the baseline profile.
The goal is to add a benchmark module and set up baseline profile generator.

- https://developer.android.com/topic/performance/baselineprofiles/create-baselineprofile

l2hyunwoo:
## Issue
- close #719 

## Overview (Required)
- Add Benchmark Module and baseline profile generator
- There's an [issue](https://github.com/DroidKaigi/conference-app-2023/issues/719#issuecomment-1697356674)


app-android/src/devRelease/generated/baselineProfiles/baseline-prof.txt
takahirom, line 1: Sorry I'm going to create this by myself without committing so could you remove this?

l2hyunwoo: It's deleted

takahirom:
@l2hyunwoo Do you have any idea why this CI takes 1 hour? 👀

l2hyunwoo:
@takahirom It's hard to find the reason why it takes much longer than others..... Reasonable reason that I can think is there's no build cache related to baseline-profile. So if build cache is created, build time will restore I suppose.

+) One more I think is there're procedures that related to prod variant, kind of strange things....(I thought it just assembleDev..) It could increase build time much more
스크린샷 2023-08-31 오후 11 13 32

takahirom:
@l2hyunwoo 
Well, you can try to use increased build time for the GitHub Action.  🙏

PR #1013, opened 2023-08-29
Avoid necessary boxing by mutableFloatStateOf

fumiya-kume:
## Issue
- N/A

## Overview (Required)
- Avoid necessary boxing by mutableFloatStateOf
- Because, `mutableStateOf` doing the boxing even if the value is primitive type.

## Links
-


fumiya-kume:
Could you re-run the Ci after finish current tasks? since non-relational of change reason, the build was failing.

takahirom (approved):
How did you find it? Great!

PR #1014, opened 2023-08-29
Convert to data object if possible

fumiya-kume:
## Issue
- N/A

## Overview (Required)

- Convert to `data object` if possible. it will improve the debugggability


takahirom (approved):
Looks great!

PR #1018, opened 2023-08-29
Improve OSS licenses screen

### LINKED ISSUE #938: Add title for the license screen
**Idea Description**

Replace to suitable title for the OSS license screen.

## Difficulty

Currently, `app-android` module not yet setup the String resources management system which using `StringsBindings`.   
For this situation, you have to do following items.

- Setup with `StringsBindings` at `app-android` module
- Call `OssLicensesMenuActivity.setActivityTitle` with added string

You can refer [Official documents](https://github.com/google/play-services-plugins/blob/master/oss-licenses-plugin/README.md#setting-the-activity-title) as well

fumiya-kume:
## Issue
- close #938 

## Overview (Required)

- Add OssLicenseScreen to list up the libraries
- Add OssLicenseDetailScreen to show the license contents
- Parsing the generated contents from the Google's licenses plugin.

## TODO

- [x]  Replace Screen title handling
- [x] Remove Licenses position from the domain model
- [x] Refactoring library grouping logic
- [x] Align naming rule for the license/library
- [x] Add implementation for navigation up
- [x] Improve Card expand state handling
- [ ] Support Large screen devices

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:
<video src="" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/16269075/64e1f7f6-7079-49f1-8afa-0afe8e72c093" width="300" >

### DIFF
--- a/core/data/src/androidMain/kotlin/io/github/droidkaigi/confsched2023/data/di/BuildConfigProviderModule.kt
+++ b/core/data/src/androidMain/kotlin/io/github/droidkaigi/confsched2023/data/di/BuildConfigProviderModule.kt
@@ -7,6 +7,7 @@ import dagger.hilt.InstallIn
 import dagger.hilt.components.SingletonComponent
 import io.github.droidkaigi.confsched2023.data.osslicense.OssLicenseDataSource
 import io.github.droidkaigi.confsched2023.model.BuildConfigProvider
+import io.github.droidkaigi.confsched2023.model.License
 import io.github.droidkaigi.confsched2023.model.OssLicenseGroup
 import kotlinx.collections.immutable.PersistentList
 import kotlinx.collections.immutable.persistentListOf
@@ -67,6 +68,17 @@ private object EmptyBuildConfigProvider : BuildConfigProvider {
 
 private object EmptyOssLicenseDataSource : OssLicenseDataSource {
     override suspend fun license(): PersistentList<OssLicenseGroup> {
-        return persistentListOf()
+        return persistentListOf(
+            OssLicenseGroup(
+                "dummy",
+                listOf(
+                    License(
+                        "id",
+                        "name",
+                        "license text",
+                    ),
+                ),
+            ),
+        )
     }
 }


app-android/src/main/java/io/github/droidkaigi/confsched2023/license/License.kt
```diff
@@ -0,0 +1,21 @@
+package io.github.droidkaigi.confsched2023.license
+
+import kotlinx.collections.immutable.PersistentList
+import kotlinx.collections.immutable.persistentListOf
+
+data class License(
```
takahirom, line 6: Could you put sources to appropriate modules?
I think 
UI will be feature/about, 
Repository's implementation and api will be core/data,
This(License) will be core/model.

fumiya-kume: That's make sense.

`OssLicenseRepositoryImpl` is still in the app module since limitation of Google's license plugin.
Except that, placing to appropriate modules. 🙇

fumiya-kume:
(Fixing the DI config :loading:

app-android/src/main/java/io/github/droidkaigi/confsched2023/license/OssLicenseRepositoryImpl.kt
```diff
@@ -0,0 +1,69 @@
...
+import kotlinx.collections.immutable.persistentListOf
+import kotlinx.collections.immutable.toPersistentList
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+import okio.BufferedSource
+import okio.buffer
+import okio.source
+import javax.inject.Inject
+
+class OssLicenseRepositoryImpl @Inject constructor(
```
takahirom, line 17: How about moving the logic to DefaultOssLicenseDataSource to simplify the app's implementation? I think this is similar to an API, since both handle data parsing. Therefore, I suggest transferring the code to DefaultOssLicenseDataSource.

```kotlin
class DefaultOssLicenseDataSource @Inject constructor(
    private val context: Context,
) : OssLicenseDataSource {

    override suspend fun licenseFlow(): OssLicense {
        return withContext(context = Dispatchers.IO) {
            val details = readLicensesFile().toRowList()
            val metadata = readLicensesMetaFile().toRowList().parseToLibraryItem(
                details = details,
            )
            val groupList = metadata.distinctBy { it.name }.groupByCategory()
                .map {
                    OssLicenseGroup(
                        title = it.key,
                        licenses = it.value,
                    )
                }.toPersistentList()
            OssLicense(groupList)
        }
    }

    private fun List<License>.groupByCategory(): Map<String, List<License>> {
        val categoryList = listOf(
            "Android Support",
            "Android Datastore",
            "Android ",
            "Compose UI",
            "Compose Material3",
            "Compose ",
            "AndroidX lifecycle",
            "AndroidX ",
            "Kotlin",
            "Dagger",
            "Firebase",
            "Ktorfit",
            "okhttp",
            "ktor",
        )
        return groupBy { license ->
            categoryList.firstOrNull {
                license.name.startsWith(
                    prefix = it,
                    ignoreCase = true,
                )
            } ?: "etc"
        }
    }

    private fun readLicensesMetaFile(): BufferedSource {
        return context.resources.openRawResource(raw.third_party_license_metadata)
            .source()
            .buffer()
    }

    private fun readLicensesFile(): BufferedSource {
        return context.resources.openRawResource(raw.third_party_licenses)
            .source()
            .buffer()
    }

    private fun List<String>.parseToLibraryItem(details: List<String>): List<License> {
        return mapIndexed { index, value ->
            val (position, name) = value.split(' ', limit = 2)
            val (offset, length) = position.split(':').map { it.toInt() }
            val id = name.replace(' ', '-')
            License(
                name = name,
                id = id,
                offset = offset,
                length = length,
                detail = details[index],
            )
        }
    }

    private fun BufferedSource.toRowList(): List<String> {
        val list: MutableList<String> = mutableListOf()
        while (true) {
            val line = readUtf8Line() ?: break
            list.add(line)
        }
        return list
    }
}
```

takahirom:
Could you add a screenshot test for this screen? This is because all other screens have one, and this PR causes our test coverage to shrink. 🙇

.github/workflows/CompareScreenshot.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           workflow: UnitTest.yml
           branch: main
 
-      - run: ./gradlew compareRoborazziDebug compareRoborazziDevDebug --stacktrace -Pscreenshot
+      - run: ./gradlew compareRoborazziDebug compareRoborazziDevRelease --stacktrace -Pscreenshot
```
takahirom, line 36: What problem are you experiencing with compareRoborazziDevDebug? 👀

PR #1019, opened 2023-08-29
[iOS] Implement the calendar registration feature on the session detail

### LINKED ISSUE #887: [iOS] Add calendar registration feature on timetable detail screen
## Summary
Now we have no functionality about calendar registration sessions. Let's make it possible to register session information to a calendar.

woxtu:
## Issue
- close #887

## Overview (Required)
- This PR implements the calendar registration feature on the session detail
  - The event module is copied from the official app 2022
  - This PR doesn't provide any localizations

## Links
- https://github.com/DroidKaigi/conference-app-2022/tree/main/app-ios/Sources/Event

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/app-ios/Modules/Sources/Session/SessionViewModel.swift
+++ b/app-ios/Modules/Sources/Session/SessionViewModel.swift
@@ -6,20 +6,18 @@ import shared
 @MainActor
 final class SessionViewModel: ObservableObject {
     @Dependency(\.eventKitClient) var eventKitClient
-    @Published var isAddingToCalendarConfirming: Bool = false
     let timetableItem: TimetableItem
 
     init(timetableItem: TimetableItem) {
         self.timetableItem = timetableItem
     }
 
-    func requestAddingToCalendar() async {
+    func requestEventAccessIfNeeded() async -> Bool {
         do {
-            if try await eventKitClient.requestAccessIfNeeded() {
-                isAddingToCalendarConfirming.toggle()
-            }
+            return try await eventKitClient.requestAccessIfNeeded()
         } catch {
             print(error.localizedDescription)
+            return false
         }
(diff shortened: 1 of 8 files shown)


app-ios/Modules/Sources/Session/SessionViewModel.swift
```diff
@@ -1,10 +1,37 @@
+import Dependencies
+import Event
 import Foundation
 import shared
 
-final class SessionViewModel {
+@MainActor
+final class SessionViewModel: ObservableObject {
+    @Dependency(\.eventKitClient) var eventKitClient
+    @Published var isAddingToCalendarConfirming: Bool = false
```
ry-itto, line 9: This is nits! I think this flag represents dialog presenting state, so `@State` might be better 👀 
How about this?

ry-itto (approved):
Almost LGTM!
I left a nits comment. Please check it 🙏🏼

app-ios/Modules/Sources/Session/SessionViewModel.swift
woxtu: My understanding is that a `@State` on an `ObservableObject` doesn't update any views.

ry-itto: @woxtu 
I didn't have enough to say. My suggestion is to use `@State` in `View` and use that 🙏🏼

woxtu: The view model takes its binding?

woxtu: Okay.

Issue #1020, opened 2023-08-29
Don't use whitelist/blacklist for inclusive codebase


fumiya-kume:
**Idea Description**

As DroidKaigi [Code of conduct](https://portal.droidkaigi.jp/about/code-of-conduct) , we won't discrimination with racial connotations.

For achieve this goal, We can treat the remaining points.

- https://github.com/DroidKaigi/conference-app-2023/blob/main/config/detekt/baseline.xml

Such as `detekt` config file still contains the "word". 

**Reference images and links**

- https://github.com/detekt/detekt/pull/2778

PR #1021, opened 2023-08-29
Remove racist words for inclusive codebase

fumiya-kume:
## Issue
- #1020 

## Overview (Required)
- We can use alternative words for inclusive codebase

## Links
- https://github.com/detekt/detekt/pull/2778

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):
Thanks!

PR #1023, opened 2023-08-29
Optimize the time table performance with reduce boxing

fumiya-kume:
## Issue
- Similar to https://github.com/DroidKaigi/conference-app-2023/pull/1013

## Overview (Required)
- Improve the performance by use suitable method.
- Reduce boxing will improve the performance


takahirom (approved):
Thanks 🗳️

PR #1024, opened 2023-08-29
Remove duplicate `.gitignore` for `.DS_Store`

fumiya-kume:
## Issue
- N/A

## Overview (Required)
- Remove duplicate `.gitignore` for `.DS_Store` 
- I think ignore once is enough.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,9 +21,6 @@ local.properties
 .classpath
 .project
 
-# Windows thumbnail db
-.DS_Store
-
 # IDEA/Android Studio project files, because
 # the project can be imported from settings.gradle.kts
 *.iml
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -21,9 +21,6 @@ local.properties
 .classpath
 .project
 
-# Windows thumbnail db
-.DS_Store
```
mhidaka, line 25: Good catch! and I prefer It looks better to replace line with `Thumbs.db` instead of just deleting it. 
Could you please do it?

fumiya-kume: Gotcha, I'm not good at Windows... so much helpful for me... 🙇

fumiya-kume: Fixed in 30589d54fb12e38d85345cfe5bc398182da7c7d7

takahirom (approved):
👍

PR #1026, opened 2023-08-29
enable to display session title when bigfont / bigdisplay

### LINKED ISSUE #464: Hide session title at session detail screen when enable the BigFont/BigDisplay

ked4ma:
# ※If this is not good solution, please close :bow: 

## Issue
- close #464

## Overview (Required)
- when set font & display to BIG, there is a possibility the seesion titles are displayed with overflow
- on LargeTopAppBar, height of large title is expected not to exceed particular value.
  - https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/AppBar.kt;l=358-359?q=AppBar.kt
- and when layout, the height will considered such like following
  - `large title height (max)` - `min title height` + `scroll offset`
  - https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/AppBar.kt;l=1246-1247?q=AppBar.kt
  - because of this, large title area will be large than dislayed actually (I think)
- I set `heightIn(...)` to set max height of `ResizableText` component as ↓
  - `large title height (max)` - `min title height`
  - https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/tokens/TopAppBarLargeTokens.kt;l=26
  - https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/tokens/TopAppBarSmallTokens.kt;l=26

Before | After
:--: | :--:
 | 
 | 
 | 

### when not bigfont / bigdisplay
※it is for making sure that there is not diff when it is not bigfont/bigdisplay.
Before | After
:--: | :--:
 | 
 | 
 |

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableItemDetailScreenTopAppBar.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableItemDetailScreenTopAppBar.kt
@@ -1,5 +1,6 @@
 package io.github.droidkaigi.confsched2023.sessions.component
 
+import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.ArrowBack
@@ -105,7 +106,7 @@ private fun ResizeableText(
             }
         },
         style = styles[styleIndex],
-        modifier = Modifier.padding(end = 16.dp),
+        modifier = Modifier.padding(end = 16.dp).heightIn(max = 88.dp),
     )
 }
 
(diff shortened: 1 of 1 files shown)


takahirom (approved):
Looks better! (I'm wondering why our screenshot tests doesn't work for this)

PR #1027, opened 2023-08-30
[iOS] Add a bottom padding to the floor map

woxtu:
## Issue
- close N/A

## Overview (Required)
- The side events on the floor map are partially hidden under the footer button.
   This PR adds a bottom padding and resolves that.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


ry-itto (approved):
Good catch! Thanks!

PR #1030, opened 2023-08-30
fix image size (aspect ratio) of speaker icon on TimetableGridITem

ked4ma:
## Issue
- no issue

## Overview (Required)
- when loading speaker icon, width of icon will be expanded to max width of `Row`.
- by setting `Modifier.aspectRatio(...)`, fix the size of icon as square.
- There are snapshot diffs.
  - the image used for preview (Icons.Default.Person) is `24.dp`.
  - so, speaker icons were displayed little bit smaller than real.

## Links
- https://developer.android.com/jetpack/compose/graphics/images/customize#custom-aspect-ratio

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/74723074/1a7254f0-a94c-436f-832a-27fea6e627d5" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/74723074/f1bb7b80-f57a-45c4-bf1e-3e7841df20fe" width="300" >

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableGridItem.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableGridItem.kt
@@ -10,7 +10,6 @@ import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
-import androidx.compose.foundation.layout.aspectRatio
 import androidx.compose.foundation.layout.defaultMinSize
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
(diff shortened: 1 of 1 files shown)


takahirom:
I'm not sure what is happening 🤔 
image

ked4ma:
Hmm. Mee too 🤔
And I don't change about TimetableListItem...

ked4ma:
@takahirom 
It was my guilty :bow:

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableGridItem.kt
```diff
@@ -218,6 +218,7 @@ private fun SpeakerIcon(
         },
         contentDescription = UserIcon.asString(),
         modifier = modifier
+            .size(TimetableGridItemSizes.speakerHeight)
```
ked4ma, line 221: When `SpeakerIcon` load very small image, icon image displayed such like a "dot".
This is because `SpeakerIcon` doen't have settings about size.
If set the size of image, it can also be resolved.
※ Almost all user icon is larger than `TimetableGridItemSizes.speakerHeight`, so I use this for `size(...)`

before | after
 :--: | :--: 
 |

takahirom (approved):
Looks great! Thanks!

PR #1031, opened 2023-08-30
:recycle: Enabled transition to archived videos and slides.

Corvus400:
## Issue
- Maybe None.

## Overview (Required)
- There was no Issue or TODO written, but I implemented a transition function because nothing was happening when I clicked on the video and slides.
- The video and slides shown in the video are what happens when a dummy URL is passed.

## Movie (Optional)
Before | After
:--: | :--:
--- | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/83cac03b-b7f0-49c7-942a-0c7ab297a81f" width="300" >


takahirom (approved):
Great! We need this feature! I forgot and thank for helping me!

PR #1033, opened 2023-08-30
:wrench: Make it possible to build again.

Corvus400:
## Issue
- None.

## Overview (Required)
- For some reason, a necessary import was missing, so it was added again.🤔


takahirom:
Thanks as always!

PR #1049, opened 2023-09-02
Improve timetable list accessibility

c0tt0n-candy:
## Issue
- no issue

## Overview (Required)
- I improved the accessibility of `TimetableList`.
    - When you enable TalkBack, `startsTimeString` and `endsTimeString`  were being read separately, so now they are read as a whole.

## Links
- [Modifier.clearAndSetSemantics](https://developer.android.com/reference/kotlin/androidx/compose/ui/semantics/package-summary#(androidx.compose.ui.Modifier).clearAndSetSemantics(kotlin.Function1))

### Before
before

### After
after


takahirom (approved):
I think this is really great change!

PR #1058, opened 2023-09-02
[iOS] Set up type-safe localizations

### LINKED ISSUE #397: [iOS] Setup type safe access to L10n
## What to do
- No localization settings for this iOS app now. Please setup localization settings and setup module that can access localization strings with type safe way. (e.g. `L10n.Timetable.title` )
  - Maybe Xcode 15 way is useful. If this way isn't good, `SwiftGen` might be better.

## Design

woxtu:
## Issue
- close #397

## Overview (Required)
- This PR sets up type-safe localizations using SwiftGen

## Links
- https://developer.apple.com/documentation/xcode/localizing-package-resources
- https://github.com/SwiftGen/SwiftGen#strings

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


tkhs0604 (approved):
Thank you for preparing the foundation of localization 🇯🇵🇺🇸

PR #1060, opened 2023-09-02
[iOS] Remove trailing whitespaces

woxtu:
## Issue
- close N/A

## Overview (Required)
- This PR removes trailing whitespaces to resolve `trailing_whitespace` rule violations

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


tkhs0604 (approved):
Thank you for fixing the violation!

PR #1061, opened 2023-09-03
:sparkles: Apply stamps initial dialog

### LINKED ISSUE #331: Apply stamps initial dialog
We would like you to implement a dialog with a warning only the first time when we transition to the stamps screen

[Figma link](https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=55510-60156&mode=dev)

Corvus400:
## Issue
- close #331 

## Overview (Required)
- A dialog appears only once when the app is installed and the stamp screen is first opened.
- A new DataStore is being registered to implement this feature.
I am not sure if this is correct.

## Movie (Optional)
Before | After
:--: | :--:
---  | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/d6f0b0d0-36b5-415b-b223-92fd138f70ed" width="300" >


core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched2023/data/stamps/StampDataStore.kt
```diff
@@ -0,0 +1,42 @@
...
+        return dataStore.data
+            .catch { exception ->
+                if (exception is IOException) {
+                    emit(emptyPreferences())
+                } else {
+                    throw exception
+                }
+            }
+            .map { preferences: Preferences ->
+                preferences[DATA_STORE_STAMPS_KEY]?.toBoolean()
```
Corvus400, line 34: @takahirom @momomomo111 
I have a question.
I assumed that when the app is installed and the preferences[DATA_STORE_STAMPS_KEY] is accessed for the first time, null is returned.
However, in reality, true is always returned from the first access after installation.
I am investigating this part of the problem, but I am unable to identify which part of the code is returning true.
Any help would be appreciated.:bow:

feature/stamps/src/main/java/io/github/droidkaigi/confsched2023/stamps/StampsScreen.kt
```diff
@@ -114,3 +133,81 @@ private fun StampsScreen(
...
+// @Composable
+// fun StampsScreenDialogPreview() {
+//    KaigiTheme {
+//        Surface {
+//            StampsScreenDialog(
+//                onDismissRequest = {},
+//            )
+//        }
+//    }
+// }
```
Corvus400, line 213: For some reason, PreviewTest fails if a preview of the Dialog is included.🤔

core/data/src/androidMain/kotlin/io/github/droidkaigi/confsched2023/data/user/DataStoreModule.kt
```diff
@@ -21,6 +21,9 @@ annotation class UserDataStoreQualifier
 @Qualifier
 annotation class SessionCacheDataStoreQualifier
 
+@Qualifier
+annotation class StampDataStoreQualifier
+
```
Corvus400, line 26: I am not sure this part is really necessary.
However, by including this implementation, we are now able to achieve the intended implementation.

Corvus400:
@takahirom @momomomo111 
Hmmm...I ran the task locally that is failing in CI and there is no error.
I apologize for all the help I've been getting lately, but do you have any clues about this?
スクリーンショット 2023-09-04 20 01 58

takahirom:
👀

Corvus400:
@takahirom 
The conflicts are so severe that I cannot fix them myself, so I will recreate the PR. 🙇

takahirom:
🙇

Issue #989, opened 2023-08-28
[iOS] Rename stamps and badges to achievements


ry-itto:
**Idea Description**

iOS version of https://github.com/DroidKaigi/conference-app-2023/issues/984

**Reference images and links**

- https://github.com/DroidKaigi/conference-app-2023/issues/984

tkhs0604:
I'm gonna fix it 🙆‍♂️

PR #1071, opened 2023-09-03
[iOS] add feature flag handle for stamp

ry-itto:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- 

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


app-ios/Modules/Sources/Navigation/RootViewModel.swift
```diff
@@ -0,0 +1,26 @@
+import Dependencies
+import Foundation
+import KMPContainer
+import Model
+
+struct RootViewState: ViewModelState {
+    var isStampEnabled: LoadingState<Bool> = .initial
```
tkhs0604, line 7: [nits]
Since we have this ticket https://github.com/DroidKaigi/conference-app-2023/issues/989, it might be better to use the word "achievement" instead.

PR #1073, opened 2023-09-03
Extract kover configurations into a build-logic plugin

### LINKED ISSUE #761: Create a single coverage report by KotlinX Cover

nashcft:
## Issue
- close #761 

## Overview (Required)

Extracted kover configurations for each projects into a plugin like [kotlinx.coroutines' convention plugin](https://github.com/Kotlin/kotlinx.coroutines/blob/1ccc96890d38be52d1bdcd31af5befb54804aa05/buildSrc/src/main/kotlin/kover-conventions.gradle.kts) so that, when we add a new module, we don't have to care about applying kover plugin to it and adding `kover` declaration to report-aggregator module. 

## Links
- 

N/A

## Movie (Optional)

N/A

### DIFF
--- a/app-android/build.gradle.kts
+++ b/app-android/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
     id("droidkaigi.primitive.android.crashlytics")
     id("droidkaigi.primitive.detekt")
     id("droidkaigi.primitive.android.roborazzi")
-    id("droidkaigi.primitive.kover")
+    id("droidkaigi.primitive.kover.entrypoint")
     id("droidkaigi.primitive.android.osslicenses")
 }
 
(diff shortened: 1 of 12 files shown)


takahirom:
Can you fix the conflict? 🙏

PR #1082, opened 2023-09-04
:recycle: Added display control of the translation function.

Corvus400:
## Issue
- None.

## Overview (Required)
- Added a new setting that does not display the Description except for normal sessions since it does not contain a Description.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/de5423f4-2cbf-4e7b-8a05-94f8ce045f7c" width="300" > |


takahirom (approved):
There is a small point but your insight is amazing!

PR #1084, opened 2023-09-04
[Android] fix calculation to offset item in grid view

tkhs0604:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- The grid layout starts from wrong position, so fixed it.
- The broken layout in day2 is the existing behavior, so keeping it as it is.

## Links
- 

Day1 | Day2
:--: | :--:
 |


feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/section/TimetableGrid.kt
```diff
@@ -330,13 +335,17 @@ private data class TimetableItemLayout(
     val minutePx: Float,
     val dayToStartTime: MutableMap<DroidKaigi2023Day, Instant>,
 ) {
+    val tz = TimeZone.of("UTC+9")
     val dayStart = dayToStartTime[timetableItem.day] ?: dayStartTime
+    val gridStartTime = LocalTime(hour = 10, minute = 0)
+    val gridStart =
```
tkhs0604, line 341: [Help wanted ⚠️]
I am not familiar with how to use date time APIs so much.
I think it is quite dirty and there should be better ways, so I would like some advice to improve this calculation 🙏

tkhs0604: Ref: Initial time is 10:00 https://github.com/DroidKaigi/conference-app-2023/blob/main/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableGridHours.kt#L271

takahirom:
@tkhs0604 I wanted not to use a fixed time for the start time. So I've changed a little. Could you take a look at it?

tkhs0604:
@takahirom 
I checked your commit and LGTM, thank you for improving 🙌

PR #1086, opened 2023-09-04
enable back at contributors

furusin:
## Issue
- there is no Issue about this PR. I just found the difference.

## Overview (Required)
- made enable to back with navigation icon at contributors view
  - took the same solution as StaffScreen
  - https://github.com/DroidKaigi/conference-app-2023/blob/main/feature/staff/src/main/kotlin/io/github/droidkaigi/confsched2023/staff/StaffScreen.kt#L88

## Links
- design: https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=55032-59308&mode=design&t=BaA1ZwkaNk0QwP3n-4

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/2215210/998b15d0-9ace-48a6-aea0-883f259e47b6" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/2215210/b06bdfca-c0c5-42a0-8030-9ceaa69e5f55" width="300" >


feature/contributors/src/commonMain/kotlin/io/github/droidkaigi/confsched2023/contributors/ContributorsScreen.kt
```diff
@@ -78,6 +79,7 @@ private fun ContributorsScreen(
             null
         }
     val localLayoutDirection = LocalLayoutDirection.current
+    val lifecycleOwner = LocalLifecycleOwner.current
```
takahirom, line 82: Good catch! Thanks for finding this issue!
But we can't use this function in commonMain because this function is used by iOS 🙏

furusin: Oh, My! Ok, I'll fix it

takahirom:
Looks great! Thanks

takahirom (approved):

furusin:
@takahirom 
Thank you for your check! Please merge whenever you can.

PR #1100, opened 2023-09-06
[iOS] Fix missing targets of the build scheme

woxtu:
## Issue
- close N/A

## Overview (Required)
- This PR adds `AchievementsTests` and `StaffTests` to the build scheme and fixes missing targets

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


tkhs0604 (approved):
Thank you for adding some targets! LGTM!

PR #1102, opened 2023-09-06
:wastebasket: Removed because it was not used.

Corvus400:
## Issue
- None.

## Overview (Required)
- Removed unneeded ExampleTest since a screenshot test exists.


takahirom (approved):
Thanks!

PR #1106, opened 2023-09-06
[Android] show snackbar message

### LINKED ISSUE #1101: Snackbar's message isn't shown
**Describe the bug**
We can't see the snack bar message

**To Reproduce**
1. Enable air plain mode
2. Go to contributor screen
3. See empty snackbar 

**Expected behavior**
You can see the message

**Screenshots**
image

ked4ma:
## Issue
- close #1101

## Overview (Required)
- Now, snackbar will be shown without any messages (only "Retry" is dispalyed).
- In this PR, fix enabling to display message.
- If message and cause is passed to RuntimeException class's constructor, message of exception will be set the value of indicated variable value.
  - the default value of message on AppError's constructor is `null` ([ref](https://github.com/DroidKaigi/conference-app-2023/blob/main/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2023/model/AppError.kt#L4))
  - so, I think message will be always `null` because message is not indicated on any classes's constructor inheritting `AppError` class

## Links
- https://github.com/JetBrains/kotlin/blob/10e94f90ac1b2396d0427d9078c1fe481859cb60/libraries/stdlib/jvm/runtime/kotlin/TypeAliases.kt#L13

Before | After
:--: | :--:
 | 
 |

### DIFF
--- a/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2023/model/AppError.kt
+++ b/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2023/model/AppError.kt
@@ -1,7 +1,8 @@
 package io.github.droidkaigi.confsched2023.model
 
 public sealed class AppError : RuntimeException {
-    private constructor(message: String? = null, cause: Throwable? = null) : super(message, cause)
+    private constructor(message: String?, cause: Throwable?) : super(message, cause)
+    private constructor(cause: Throwable?) : super(cause)
 
     public sealed class ApiException(cause: Throwable?) : AppError(cause = cause) {
         public class NetworkException(cause: Throwable?) : ApiException(cause = cause)


core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2023/model/AppError.kt
```diff
@@ -1,7 +1,8 @@
 package io.github.droidkaigi.confsched2023.model
 
 public sealed class AppError : RuntimeException {
-    private constructor(message: String? = null, cause: Throwable? = null) : super(message, cause)
+    private constructor(message: String?, cause: Throwable?) : super(message, cause)
```
ked4ma, line 4: I have kept this constructor considering the possibility that we might want to set a fixed message in the future. 
Please let me know in the comments if you think it should be removed :pray:

takahirom (approved):
I've learned a lot! thanks!

PR #1107, opened 2023-09-06
[Android] fix bottom padding on contributor screen

ked4ma:
## Issue
- NO_ISSUE

## Overview (Required)
- As following image(Before), bottom element of contributor will be buried behind the navbar.
  - with some device, snackbar is not displayed due to the navbar.
- Now, contributor screen is move into about screen
  - so I tried to change padding setting same as staff and sponsor screen to enable to show bottom element and snackbar.
- This doesn't have an issue ticket, so if it's AWFUL change, please close this PR :bow: 

## Links
- nothing

device | Before | After
:--: | :--: | :--:
Pixel 6 api 30 |  | 
Pixel Fold api 33 |  |

### DIFF
--- a/feature/contributors/src/commonMain/kotlin/io/github/droidkaigi/confsched2023/contributors/ContributorsScreen.kt
+++ b/feature/contributors/src/commonMain/kotlin/io/github/droidkaigi/confsched2023/contributors/ContributorsScreen.kt
@@ -1,9 +1,8 @@
 package io.github.droidkaigi.confsched2023.contributors
 
-import androidx.compose.foundation.layout.PaddingValues
-import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.fillMaxSize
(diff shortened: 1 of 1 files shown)


ked4ma:
@takahirom 
> Our policy for inset padding is this. Do you have any idea why it doesn't work? [#959 (comment)](https://github.com/DroidKaigi/conference-app-2023/pull/959#discussion_r1306396623)

I this it's because `ContributorScreen` is not in the `mainNestedGraph` now.
In #959, contentPadding is passed from mainNestedGraph and had been using it for showing ContributorScreen.
But now, this is moved into the `KaigiNavHost` func.
And this func doesn't have padding information so ContributorScreen will use default `PaddingValues` .
By default, `PaddingValues` have `0dp` for horizontal and vertical.
As a result, I think how to do padding was changed :thinking:

takahirom (approved):

Issue #1062, opened 2023-09-03
[Compose Multiplatform iOS]Apply the app's theme explicitly by adding KaigiTheme at the root of ContributorsScreen to change colors in this content


takahirom:
**Idea Description**
![image](https://github.com/DroidKaigi/conference-app-2023/assets/1386930/c2737d80-9db3-4cb1-bbd0-be53809bd445)

**Reference images and links**
https://github.com/DroidKaigi/conference-app-2023/pull/1054#issuecomment-1703973359

ttanaka330:
Can i try this issue?

takahirom:
@ttanaka330 Thanks! Assigned 👍 
I just confirm that this may involve iOS app build. Is it ok?

ttanaka330:
@takahirom 
Sorry, I just cannot confirm that this may involve iOS app build.

takahirom:
@ttanaka330 It's ok! Thanks! I've edited the title 🙏

tkhs0604:
📝 KaigiTheme refers to the custom font file, so the following error occurs when just wrapping ContributesScreen by KaigiTheme:
> Uncaught Kotlin exception: kotlin.IllegalArgumentException: org.jetbrains.compose.resources.MissingResourceException: Missing resource with path: font/montserrat_medium.ttf

If KaigiTheme does not refer to any font files, that is, the default font is used, it seems to work well.
<img src="https://github.com/DroidKaigi/conference-app-2023/assets/43767445/35547fe2-82e9-42a1-abce-4bfd0e938ebd" width="300">

PR #1108, opened 2023-09-06
[Compose Multiplatform] apply theme to contributors screen (w/ workaround)

tkhs0604:
## Issue
- close https://github.com/DroidKaigi/conference-app-2023/issues/1062

## Overview (Required)
- Apply KaigiTheme to contributions screen for iOS
- There is an issue that the custom font cannot be read from iOS side, so fixed to return the system font as a workaround 🙏

## Links
-


takahirom (approved):
Done is better than perfect! Thanks

PR #1113, opened 2023-09-07
[Android] Enable to get version name from BuildConfig on AboutScreen

syarihu:
## Issue
- No issue

## Overview (Required)
- Enable to get the version name from BuildConfig on AboutScreen. This eliminates the need to get the version name from Context

### Detail
- Create a BuildConfigProvider interface to enable provision of BuildConfig values
- Create an AppModule to get the app version name from BuildConfig and pass it to BuildConfigProvider
- Added version name to AboutScreenUiState, which was unused, and made UiState available
- Added process to ViewModel to get app version name from BuildConfigProvider and set it to UiState
- Replaced the part of getting the app version name from Context with getting it from UiState
- The function to get the version name from Context is no longer needed, so it was removed

## Links
- Nothing

- Ensure results remain the same

Before | After
:--: | :--:
 |


takahirom:
We've attempted this. However, there's an issue where the tests can't recognize the Dagger Module of the app-android module. I believe the tests will fail. 😭 
https://github.com/DroidKaigi/conference-app-2023/pull/782

syarihu:
@takahirom 
I have created a BuildConfigProviderModule that provides an empty BuildConfigProvider that can override the Provider using BindsOptionalOf in the data module.
Then, by overriding that setting in the AppModule, app-android was able to return values from BuildConfig while providing an empty BuildConfigProvider in the test.
I think this also makes all the test code PASS. In my environment, I passed all of them.
How about this approach?

feature/about/src/main/java/io/github/droidkaigi/confsched2023/about/AboutScreenViewModel.kt
```diff
@@ -2,18 +2,22 @@ package io.github.droidkaigi.confsched2023.about
...
     val userMessageStateHolder: UserMessageStateHolder,
+    buildConfigProvider: BuildConfigProvider,
 ) : ViewModel(),
     UserMessageStateHolder by userMessageStateHolder {
+    private val versionNameStateFlow = MutableStateFlow(buildConfigProvider.versionName)
 
-    val uiState = MutableStateFlow(
-        AboutScreenUiState(),
-    ).asStateFlow()
+    val uiState: StateFlow<AboutScreenUiState> = buildUiState(versionNameStateFlow) {
```
takahirom, line 20: 👍

takahirom:
@matsumo0922  For your reference, the issue is fixed now! 👍

takahirom (approved):
I have never used the BindsOptionalOf! Thanks

syarihu:
Thanks for your review and merge!

Issue #1117, opened 2023-09-08
The location of each event on the third day of Droidkaigi is wrong.


Corvus400:
**Describe the bug**
- Each event on the 3rd day of Droidkaigi is supposed to be held on the 1st floor as shown in the attached image, but perhaps the data structure does not support it and it is displayed as the 1st floor of the basement.
- I would like to fix it to show the 1st floor at least.
- If possible, I would like to be able to display which room is on the first floor.

**To Reproduce**
- Open any event on the third day.

**Expected behavior**

- Better: indicated as 1F.
- Best: The room name is displayed according to the room of Day2,3.

**Screenshots**

<img src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/db55ce0a-d1af-4012-9253-bc570626e0cd" width="300">
<img src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/7fdad541-c9fe-4014-bce4-5b161c0c81da" width="300">
<img src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/9761b910-231f-4f72-9b57-960dd9194889" width="300">

Corvus400:
I would like to do this problem. 🙋

Corvus400:
@takahirom 
Let me ask you a question.🙋
I know we will be using rooms on Day 3 except for the office tour, but which room on the floor map is the room in the attached image?
Or should the floor map image be different for Day 3 and other days? (In that case, we would need a new image for the floor map, which would put a burden on the designer.)

![スクリーンショット 2023-09-08 17 49 31](https://github.com/DroidKaigi/conference-app-2023/assets/13657682/85ebaebf-6bda-4608-84bb-90117ff718ed)

takahirom:
Thanks! As far as I know, all the events will take place in one room, so you don't need to worry about finding different rooms. (Note: This information is unofficial and subject to change or correction.)

takahirom:
But about the label of the detail screen. We might have to consider something 💭 It is not a room actually.

Corvus400:
@takahirom 
I see...I guess I shouldn't step in and deal with this until the specifications for how to display the information are decided a bit.
However, I would like to fix the first basement floor, so I will send out a PR to fix that.🫡

PR #1120, opened 2023-09-08
:wrench: Correctly indicated that the third day room is on the first floor, not the basement.

Corvus400:
## Issue
- #1117 
- CLOSES are not attached, as this will be a partial modification of the ticket content.

## Overview (Required)
- The room on the third day will be the room where D and E are combined. 

## Links
- https://twitter.com/mhidaka/status/1700079778643198299

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/da7c8e2e-9d95-4c96-a3b3-3e0ee1e06eae" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/0d5d9a08-769d-42fc-842a-f79aef264c55" width="300" >

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableRoom.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableRoom.kt
@@ -7,6 +7,7 @@ import io.github.droidkaigi.confsched2023.model.RoomType.RoomA
 import io.github.droidkaigi.confsched2023.model.RoomType.RoomB
 import io.github.droidkaigi.confsched2023.model.RoomType.RoomC
 import io.github.droidkaigi.confsched2023.model.RoomType.RoomD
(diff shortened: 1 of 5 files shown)


feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/TimetableRoom.kt
```diff
@@ -20,6 +21,8 @@ internal val TimetableRoom.color: Color
             RoomC -> colors.hallC
             RoomD -> colors.hallD
             RoomE -> colors.hallE
+            // The color of D is set as workaround.
+            RoomDE -> colors.hallD
```
Corvus400, line 25: I would like to assign a room color for the third day or a gradient color for D and E if possible, but I have decided that it is not advisable to make major changes to this structure now.

takahirom (approved):
I think this is better than what we have now! Thanks!

PR #1121, opened 2023-09-08
:recycle: A summary statement of the third day's content is now displayed.

Corvus400:
## Issue
- None.

## Overview (Required)
- The website shows a summary statement of the third day's content, but the app does not currently show it.
- Therefore, we have modified it so that it is displayed.
- In addition, the Read More button is no longer displayed when the text is short.

## Links
- https://2023.droidkaigi.jp/timetable/e2d76d8b-f875-4919-810c-62a9b84b82df/

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/61a25bb5-757f-4bcf-8727-c4d0e5a32cf5" width="300" > |


takahirom (approved):
We almost forgot to provide the important information! Thanks for catching that!

PR #1123, opened 2023-09-08
:recycle: Day3 shows the 1st floor and the rest show the basement.

Corvus400:
## Issue
- None.

## Overview (Required)
- The number of floors displayed by default was changed to default to the first floor only on the third day, since Day 3 does not go to the basement.

## Movie (Optional)
Third Day | Other Day
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/0ba703db-7eaa-4e87-aa93-7868875f31b4" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/cfabd793-ecdd-4455-b2d9-3c656f5881f5" width="300" >


momomomo111 (approved):
LGTM!

PR #1125, opened 2023-09-08
chore - remove isTest in SearchTextFieldAppBar Method

### LINKED ISSUE #666: PreviewTest hang when you use `focusRequester.requestFocus()`
**Idea Description**
Currently, we are checking if this is a test for preventing hanging the test.
In this issue, please investigate the reason why the test hangs and how we should treat them.

https://github.com/DroidKaigi/conference-app-2023/blob/5a13fb7f988080148633b88c2549c3ab47bae7be/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/SearchTextFieldAppBar.kt#L83-L85

Nanamare:
## Issue
- close #666

## Overview (Required)
- All UnitTests appear to succeed. I think it has been resolved naturally now.

## Links
- 

스크린샷 2023-09-09 오전 2 53 44


takahirom:
May I ask if you checked PreviewTest?

PR #1126, opened 2023-09-08
feat - add haptic feedback

### LINKED ISSUE #1056: Haptic feedback for session bookmarking
**Idea Description**
Haptic feedback for session bookmarking is an idea from the designer. It might be a great feature, but I don't want it to become annoying. Can we implement it in a subtle way?

Nanamare:
## Issue
- close #1056

## Overview (Required)
- Appropriate haptic feedback was processed 🚀 

## Links
-

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/section/TimetableList.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/section/TimetableList.kt
@@ -84,14 +84,17 @@ fun TimetableList(
                 }
             }
             Row(
-                modifier = Modifier.padding(start = 16.dp).onGloballyPositioned {
-                    rowHeight = it.size.height
-                },
+                modifier = Modifier
+                    .padding(start = 16.dp)
+                    .onGloballyPositioned {
+                        rowHeight = it.size.height
+                    },
             ) {
                 Column(
                     modifier = Modifier
                         .padding(top = 10.dp)
-                        .width(58.dp).onGloballyPositioned {
+                        .width(58.dp)
+                        .onGloballyPositioned {
                             timeTextHeight = it.size.height
                         }
                         .offset { IntOffset(0, timeTextOffset) },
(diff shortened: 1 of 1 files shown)


takahirom:
As you can see, there is a person who raises a hand in the issue first. So if the person doesn't make a PR, I will merge this PR 🙇

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/section/TimetableList.kt
```diff
@@ -128,13 +134,21 @@ fun TimetableList(
...
                             timetableItem,
                             modifier = Modifier
                                 .clickable { onTimetableItemClick(timetableItem) }
                                 .padding(top = 10.dp),
                             isBookmarked = isBookmarked,
-                            onBookmarkClick = onBookmarkClick,
+                            onBookmarkClick = { timetableItem, isBookmarked ->
+                                haptic.performHapticFeedback(
+                                    HapticFeedbackType(
+                                        HapticFeedbackConstants.LONG_PRESS,
```
takahirom, line 147: May I ask why you used LONG_PRESS here? 🙋

Nanamare: @takahirom 
it was supported starting from API Level 3, and the user experience and strength of haptic feedback were general! and It looks like it's long because it written Long_PRESS, but it's often used for general touch as well.

Do you have any better ideas ?!

takahirom: I'm just not familiar with this API and I was just wondering because this click is not a long press. Is there any API we can use for this? 👀

Nanamare: How about [VIRTUAL_KEY](https://developer.android.com/reference/android/view/HapticFeedbackConstants#VIRTUAL_KEY
) or [KEYBOARD_TAP](https://developer.android.com/reference/android/view/HapticFeedbackConstants#KEYBOARD_TAP) ?!

takahirom: @kosenda @fumiya-kume 
May I ask your opinion on this?

kosenda: Personally, I think `LONG_PRESS` is fine. 
Also, this may be off topic, but I thought Haptics was unnecessary when removing bookmarks.

Nanamare: Oh, that's a good idea.
4330d478f1f7715c612d7d2f0c8b520c9ee4e54e

takahirom (approved):
Thanks for applying the feedback. Let's see what it is like!

PR #1127, opened 2023-09-08
[Proposal] :recycle: Added the ability to go to the floor map from the session details.

Corvus400:
## Issue
- None.

## Overview (Required)
- Added the ability to go to the floor map from the session details.
- I really wanted to have the display switch between the first floor and the basement according to the floor level of each room, but I gave up on that this time.

## Movie (Optional)
Before | After
:--: | :--:
<video src="" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/bae1e487-cea9-424f-9d05-b0b99b7d13b4" width="300" >

### DIFF
--- a/app-android/src/main/java/io/github/droidkaigi/confsched2023/AppModule.kt
+++ b/app-android/src/main/java/io/github/droidkaigi/confsched2023/AppModule.kt
@@ -6,6 +6,9 @@ import dagger.hilt.InstallIn
 import dagger.hilt.components.SingletonComponent
 import io.github.droidkaigi.confsched2023.data.di.AppAndroidBuildConfig
 import io.github.droidkaigi.confsched2023.model.BuildConfigProvider
+import io.github.droidkaigi.confsched2023.model.NavigationRequester
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
 import javax.inject.Singleton
 
 @InstallIn(SingletonComponent::class)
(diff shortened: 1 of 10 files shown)


app-android/src/main/java/io/github/droidkaigi/confsched2023/AppModule.kt
```diff
@@ -15,8 +18,25 @@ class AppModule {
...
+    override fun getNavigationRouteFlow(): Flow<String> = _navigateRequestStateFlow
+
+    override fun navigateTo(route: String) {
+        _navigateRequestStateFlow.value = route
+    }
+
+    override fun navigated() {
+        _navigateRequestStateFlow.value = ""
+    }
+}
```
Corvus400, line 42: We did this to operate between different NavHosts.
I don't think this is the best way to do it, so if you have a better idea, I'm open to it.

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableItemDetailScreen.kt
```diff
@@ -109,6 +112,10 @@ fun TimetableItemDetailScreen(
         onNavigationIconClick = onNavigationIconClick,
         onBookmarkClick = viewModel::onBookmarkClick,
         onLinkClick = onLinkClick,
+        onRoomClick = {
+            viewModel.navigateTo("floorMap")
```
Corvus400, line 116: I would like to change this part if possible since it is directly specified.

app-android/src/main/java/io/github/droidkaigi/confsched2023/KaigiApp.kt
```diff
@@ -111,6 +111,9 @@ private fun KaigiNavHost(
             onTimetableItemClick = navController::navigateToTimetableItemDetailScreen,
             onNavigateToBookmarkScreenRequested = navController::navigateToBookmarkScreen,
             onLinkClick = externalNavController::navigate,
+            onRoomClick = {
+                navController.popBackStack(navController.graph.startDestinationId, false)
```
takahirom, line 115: Well, it's difficult to say but what if we create `navController::navigateToMainFloorMapScreen`?
I think we can see parent navController to access getBackStackEntry 👀

Corvus400: Once I was able to return to the start destination with the following code.

```kotlin
fun NavController.navigateToMainFloorMapScreen() {
    getBackStackEntry(timetableItemDetailScreenRoute).destination.parent?.startDestinationId?.let {
        navigate(it)
        // TODO navigateMainFloorMap
    }
}
```

However, I have not figured out how to move to FloorMapScreen from here without having to make any major changes.
Do you already have any ideas?

takahirom: I'm not sure about this. But I think we can pass the `route` parameter to NavHost and we may be able to pass /main/floorMap.

Corvus400: @takahirom 
I have tried the suggested method, but unfortunately the app crashes. 🙇 
We still think that as long as NavHost is split into two, it is unfortunately difficult to achieve this functionality other than this current implementation. 🙇

PR #1131, opened 2023-09-09
Adjust Timetable App Bar Bookmark Icon State

### LINKED ISSUE #1116: [Android] Unpredictable Timetable App Bar Bookmark Icon State
**Describe the bug**
Reproduced with an app released from the Play Store.
The state of the bookmark icon in the top app bar resets upon restarting the app. I'm not sure if this is the intended behavior or a bug.

**To Reproduce**
Follow these steps to reproduce the behavior:

1. Verify the visibility of the bookmark list icon in the top app bar.
2. Tap the 'bookmark icon' found on the timetable within a session item.
3. Observe that the bookmark list icon in the top app bar changes (it fills with color and a '+' appears on the icon).
4. Restart the app.
5. Notice that the bookmark list icon in the top app bar reverts to its state from step 2.

Sometimes didn't happen, but you can see the other pattern's state management problems.

**Expected behavior**
In this scenario, the state of the bookmark list icon in the top app bar should not change when the app is restarted.

**Screenshots**

I will attach the recorded one later. 🙏

ked4ma:
## Issue
- close #1116

## Overview (Required)
- bookmark icon has unexpected behaviors at app bar on timetable screen (refer `before` movie)
- HOT to REPRODUCTION
  1. add bookmark on timetable list screen
  2. go to bookmark screen
  3. go back to timetable list screen
- WHY the behavior like this
  - ■ when simply adding bookmark
    <details>
      <summary>expand for detail...</summary>

      | onBookmarkClickState | isPlaying | progress | remarks |
      | :--: | :--: | :--: | :--: |
      NULL | FALSE |   | [[at first]]
      TRUE | FALSE |   | when click bookmark icon
      TRUE | TRUE | 0 | start animation
      TRUE | TRUE | ~1.0 |  
      TRUE | FALSE | 1 | call onReachAnimationEnd()
      FALSE | FALSE | 1 |  
      FALSE | TRUE | 1 | isPlay will be TRUE because onBookmarkClickState is updated
      FALSE | FALSE | 0 |  
      TRUE | FALSE | 0 | [[second click]]
      TRUE | TRUE | 0 | start animation
      TRUE | TRUE | ~1.0 |  
      TRUE | FALSE | 1 | call onReachAnimationEnd()
      FALSE | FALSE | 1 |  
      FALSE | TRUE | 1 | isPlay will be TRUE because onBookmarkClickState is updated
      FALSE | FALSE | 0 |  
    </details>

  - ■ when go back from bookmark list (before going to bookmark list, adding bookmark)
    - ※1,2): operated in one composable execution
    - the problem here is 
      - **onBookmarkClickState is false at first**
      - **onBookmarkClickState is false right after click bookmark**
    <details>
      <summary>expand for detail...</summary>

      onBookmarkClickState | isPlaying | progress | remarks
      -- | -- | -- | --
      FALSE | TRUE | 0 | animation is started because onBookmarkClickState is NOT null
      FALSE | TRUE | ~1.0 |  
      FALSE | FALSE | 1 | onReachAnimationEnd()
      **isPlaying will not be TRUE again because onBookmarkClickState is FALSE at first**
      TRUE | FALSE | 1 | [[second click]]
      TRUE※1 | TRUE※1 | 1 | isPlaying become TRUE (∵onBookmarkClickState is TRUE)
      FALSE※2 | FALSE※2 | 1 | onBookmarkClickState will be updated to FALSE becuase progress is 1.0
      FALSE | TRUE | 0 | start animation becuase onBookmarkState is updated
      FALSE | TRUE | ~1.0 |  
      FALSE | FALSE | 1 | onReachAnimationEnd()
      **isPlaying will not be TRUE again because onBookmarkClickState is FALSE (unchanged)** | | | |
    </details>
- I fix this issue by changing how to judge "animation is finished"
  - use LottieAnimationState (isPlaying, isAtEnd, progress) instead of using only `progress`
  - as a result, first null value of onBookmarkClickState and animationRange could remove
  - by this, variables arround animation are changed as following
    - ■ when simply adding bookmark
      <details>
        <summary>expand for detail...</summary>

        onBookmarkClickState | isPlaying | progress(state) | isPlaying(state) | isAtEnd(state) | remarks
        -- | -- | -- | -- | -- | --
        FALSE | FALSE | 0 | FALSE | TRUE | [[at first]]
        TRUE | FALSE | 0 | FALSE | TRUE | when click bookmark icon
        TRUE | TRUE | 0 | TRUE | FALSE | start animation
        TRUE | TRUE | ~1.0 | TRUE | FALSE |  
        TRUE | TRUE | 1 | TRUE | TRUE | when animation finished
        FALSE | FALSE | 1 | FALSE | TRUE | animation state is updated according to isPlaying
        TRUE | FALSE | 1 | FALSE | TRUE | [[second click]]
        TRUE | TRUE | 0 | TRUE | FALSE | start animation
        TRUE | TRUE | ~1.0 | TRUE | FALSE |  
        TRUE | TRUE | 1 | TRUE | TRUE | when animation finished
        FALSE | FALSE | 1 | FALSE | TRUE | animation state is updated according to isPlaying
      </details>
    - ■ when go back from bookmark list (before going to bookmark list, adding bookmark)
      <details>
        <summary>expand for detail...</summary>

        onBookmarkClickState | isPlaying | progress(state) | isPlaying(state) | isAtEnd(state) | remarks
        -- | -- | -- | -- | -- | --
        FALSE | FALSE | 0 | FALSE | TRUE | animation won't start because isPlaying will be TRUE only when onBookmarkClickState is TRUE
      </details>
## Links
- nothing

## Movie
Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/74723074/da71fc7c-51c1-4c00-8de2-cca035230f96" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/74723074/030bf26b-0638-41a3-bb19-26f02ce7073e" width="300" >

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableScreen.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableScreen.kt
@@ -114,7 +114,7 @@ fun TimetableScreen(
 data class TimetableScreenUiState(
     val contentUiState: TimetableSheetUiState,
     val timetableUiType: TimetableUiType,
-    val onBookmarkIconClickStatus: Boolean?,
+    val onBookmarkIconClickStatus: Boolean,
 )
 
 private val timetableTopBackgroundLight = Color(0xFFF6FFD3)
@@ -257,7 +257,7 @@ fun PreviewTimetableScreenDark() {
                     ),
                 ),
                 TimetableUiType.Grid,
-                null,
+                false,
             ),
             SnackbarHostState(),
             {},
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableScreen.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableScreen.kt
@@ -247,8 +247,8 @@ private fun TimetableScreen(
 fun PreviewTimetableScreenDark() {
     KaigiTheme {
         TimetableScreen(
-            TimetableScreenUiState(
-                TimetableSheetUiState.ListTimetable(
+            uiState = TimetableScreenUiState(
+                contentUiState = TimetableSheetUiState.ListTimetable(
                     mapOf(
                         DroidKaigi2023Day.Day1 to TimetableListUiState(
                             mapOf<String, List<TimetableItem>>().toPersistentMap(),
@@ -256,17 +256,17 @@ fun PreviewTimetableScreenDark() {
                         ),
                     ),
                 ),
-                TimetableUiType.Grid,
-                false,
+                timetableUiType = TimetableUiType.Grid,
+                onBookmarkIconClickStatus = false,
             ),
-            SnackbarHostState(),
-            {},
-            { _, _ -> },
-            {},
-            {},
-            {},
-            {},
-            Modifier.statusBarsPadding(),
+            snackbarHostState = SnackbarHostState(),
+            onTimetableItemClick = {},
+            onBookmarkClick = { _, _ -> },
+            onBookmarkIconClick = {},
+            onSearchClick = {},
+            onTimetableUiChangeClick = {},
+            onReachAnimationEnd = {},
+            modifier = Modifier.statusBarsPadding(),
         )
     }
 }
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/BookmarkIcon.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/BookmarkIcon.kt
@@ -32,7 +32,7 @@ fun BookmarkIcon(
     val lottieComposition by rememberLottieComposition(RawRes(raw.add_to_bookmark_lottie))
     var isPlaying by remember { mutableStateOf(false) }
 
-    val state = animateLottieCompositionAsState(
+    val lottieState = animateLottieCompositionAsState(
         composition = lottieComposition,
         isPlaying = isPlaying,
         restartOnPlay = true,
@@ -44,7 +44,7 @@ fun BookmarkIcon(
         }
     }
 
-    if (state.isPlaying && state.isAtEnd) {
+    if (lottieState.isPlaying && lottieState.isAtEnd) {
         isPlaying = false
         onReachAnimationEnd()
     }
@@ -53,10 +53,10 @@ fun BookmarkIcon(
         modifier = modifier,
         contentAlignment = Alignment.Center,
     ) {
-        if (state.isPlaying && !state.isAtEnd) {
+        if (lottieState.isPlaying && !lottieState.isAtEnd) {
             LottieAnimation(
                 composition = lottieComposition,
-                progress = { state.progress },
+                progress = { lottieState.progress },
                 modifier = Modifier
                     .semantics {
                         onClick(label = contentDescription, action = null)
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/BookmarkIcon.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/BookmarkIcon.kt
@@ -17,7 +17,6 @@ import androidx.compose.ui.semantics.onClick
 import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.unit.dp
 import com.airbnb.lottie.compose.LottieAnimation
-import com.airbnb.lottie.compose.LottieCancellationBehavior
 import com.airbnb.lottie.compose.LottieCompositionSpec.RawRes
 import com.airbnb.lottie.compose.animateLottieCompositionAsState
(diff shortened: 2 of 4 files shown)


feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/BookmarkIcon.kt
```diff
@@ -25,27 +25,26 @@ import io.github.droidkaigi.confsched2023.feature.sessions.R.raw
...
-    val progress by animateLottieCompositionAsState(
+    val state = animateLottieCompositionAsState(
         composition = lottieComposition,
         isPlaying = isPlaying,
         restartOnPlay = true,
     )
 
     LaunchedEffect(onBookmarkClickStatus) {
-        if (onBookmarkClickStatus != null) {
+        if (onBookmarkClickStatus) {
```
ked4ma, line 42: update `isPlaying` to true only when `onBookmarkClickStatus` is `true`.
this is for not launcing animation when `onBookmarkClickState` is `false`.

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/BookmarkIcon.kt
```diff
@@ -25,27 +25,26 @@ import io.github.droidkaigi.confsched2023.feature.sessions.R.raw
...
             isPlaying = true
         }
     }
 
-    if (progress == 1f) {
+    if (state.isPlaying && state.isAtEnd) {
```
ked4ma, line 47: instead of judge animation ending by `progress`,
use `isPlaying` nad `isAtEnd` parameter of `LottieAnimationState`.

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/BookmarkIcon.kt
```diff
@@ -54,10 +53,10 @@ fun BookmarkIcon(
         modifier = modifier,
         contentAlignment = Alignment.Center,
     ) {
-        if (progress in animationRange) {
+        if (state.isPlaying && !state.isAtEnd) {
```
ked4ma, line 56: same as [L47](https://github.com/DroidKaigi/conference-app-2023/pull/1131/files#diff-5ab60002983aa4756e673f60b4537efd96e32a3ef541d84dd60778115ba16c28R47), we can judge whether animation is playing or not by `LottieAnimationState`

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableScreen.kt
```diff
@@ -257,7 +257,7 @@ fun PreviewTimetableScreenDark() {
                     ),
                 ),
                 TimetableUiType.Grid,
-                null,
+                false,
```
takahirom, line 260: This is an existing implementation, but it is difficult to review because I can only see a change from null to false. Could you add a named parameter for this? 🙇

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/BookmarkIcon.kt
```diff
@@ -25,27 +25,26 @@ import io.github.droidkaigi.confsched2023.feature.sessions.R.raw
...
+    onBookmarkClickStatus: Boolean,
     onReachAnimationEnd: () -> Unit,
     modifier: Modifier = Modifier,
 ) {
     val lottieComposition by rememberLottieComposition(RawRes(raw.add_to_bookmark_lottie))
-    val animationRange = 0.1f..1f
     var isPlaying by remember { mutableStateOf(false) }
 
-    val progress by animateLottieCompositionAsState(
+    val state = animateLottieCompositionAsState(
```
takahirom, line 35: What if we name this `lottieCompositionState`?

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/TimetableScreen.kt
ked4ma: I add parameter names to this preview :smiley: 
(https://github.com/DroidKaigi/conference-app-2023/pull/1131/commits/66f1cce05a15f6a39a2049ca2096fec77048a286)

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/component/BookmarkIcon.kt
ked4ma: it's more understandable.
i changed it from `state` to `lottieState`
(https://github.com/DroidKaigi/conference-app-2023/pull/1131/commits/5e7f7cf99450149105ee10aadd5a4773fce37cb1)

takahirom (approved):
Looks great! Thanks for the consistent screen!

PR #1133, opened 2023-09-09
Empty tags may appear in BookmarkList and SearchList (in code) 

### LINKED ISSUE #1069: Empty tags may appear in BookmarkList and SearchList (in code)
**Idea Description**
Empty tags may appear in BookmarkList and SearchList (in code)
In the process of creating tags for each item in the bookmark list and search results list, we are adding an empty tag if `timetableItem.day?.name` is null. This has led to the possibility of empty string tags being created and displayed in the code.

- in BookmarkList
https://github.com/DroidKaigi/conference-app-2023/blob/8ac1aa8273440566cc20665d8e6dfe4e3cda3eef/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/section/BookmarkList.kt#L119

- in SearchList
https://github.com/DroidKaigi/conference-app-2023/blob/8ac1aa8273440566cc20665d8e6dfe4e3cda3eef/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/section/SearchList.kt#L141

Due to the nature of the product, it is unlikely that `timetableItem.day?.name` would be null. Instead of using `timetableItem.day?.name.orEmpty()`, I believe it would be more appropriate to use `timetableItem.day?.name!!` or only create the tag if `timetableItem.day?.name` is not null.

kkkkan:
## Issue
- close #1069 

## Overview (Required)
- When creating BookmarkList and SearchList tags, we now check that day is not null before creating them.

## Links
none

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/section/BookmarkList.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/section/BookmarkList.kt
@@ -115,10 +115,13 @@ fun BookmarkList(
                                     labelColor = roomChipLabelColor,
                                     backgroundColor = roomChipBackgroundColor,
                                 )
-                                SessionTag(
-                                    label = timetableItem.day?.name.orEmpty(),
-                                    borderColor = MaterialTheme.colorScheme.outline,
-                                )
+                                timetableItem.day?.name?.let {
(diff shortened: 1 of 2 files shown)


takahirom (approved):
Looks great! Thanks

PR #1134, opened 2023-09-09
Changing the fetch timing of RemotoConfig

momomomo111:
## Overview (Required)
- I changed the update timing in RemoteConfig to the app's first launch and when the app is restarted.
  - It has been confirmed that updates are made through RemoteConfig upon app restart.


core/data/src/androidMain/kotlin/io/github/droidkaigi/confsched2023/data/remoteconfig/DefaultRemoteConfigApi.kt
```diff
@@ -4,21 +4,20 @@ import co.touchlab.kermit.Logger
...
-//                minimumFetchIntervalInSeconds = 1 * 60
-//            }
-//        }
-//    }
+    init {
+        CoroutineScope(Dispatchers.IO).launch {
+            firebaseRemoteConfig.reset()
+            fetchConfig()
+        }
+    }
```
momomomo111, line 20: The app updates when it is restarted.

takahirom:
Let me close this

PR #1135, opened 2023-09-09
[iOS] :sparkles: Typography was defined in Swift as in Android.

### LINKED ISSUE #889: [iOS] Add typography set as a theme
## Summary
We need to add typography set into theme module.
https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=56041-40580&mode=design&t=INXVQMdn5XadnO7e-0

Corvus400:
## Issue
- close #889 

## Overview (Required)
- The title says it all.


app-ios/Modules/Sources/Theme/Typography.swift
```diff
@@ -0,0 +1,47 @@
...
+    var fontFamily: String
+    var fontWeight: SwiftUI.Font.Weight
+    var fontStyle: SwiftUI.Font
+    var lineHeight: CGFloat
+    var letterSpacing: CGFloat?
+}
+
+struct TypographyTokens {
+    
+    static let DisplayLarge = TextStyle(fontSize: 57, fontFamily: "Montserrat-Regular", fontWeight: .regular, fontStyle: SwiftUI.Font.custom("Montserrat-Regular", size: 57), lineHeight: 64)
```
ry-itto, line 32: How about this? 👀 
I think type safe access to resource is better way 👍🏼 

Like this.
```suggestion
struct TextStyle {
    var font: Font
    var lineHeight: CGFloat
    var letterSpacing: CGFloat?
}

struct TypographyTokens {
    static let DisplayLarge = TextStyle(font: FontAssets.Montserrat.regular.font(size: 57), lineHeight: 64)

```

app-ios/Modules/Sources/Theme/Typography.swift
```diff
@@ -0,0 +1,47 @@
...
+    static let DisplayMedium = TextStyle(fontSize: 45, fontFamily: "Montserrat-Regular", fontWeight: .regular, fontStyle: SwiftUI.Font.custom("Montserrat-Regular", size: 45), lineHeight: 52)
```
ry-itto, line 33: Generally, static variables in Swift are defined in lower camel case. Can you change these? 👀

app-ios/Modules/Sources/Theme/Typography.swift
```diff
@@ -0,0 +1,47 @@
...
+}
+
+struct TextStyle {
+    var fontSize: CGFloat
+    var fontFamily: String
+    var fontWeight: SwiftUI.Font.Weight
+    var fontStyle: SwiftUI.Font
+    var lineHeight: CGFloat
+    var letterSpacing: CGFloat?
+}
```
ry-itto, line 28: I think these struct and variables should be `public` accessible level for access from other modules 👀 
If my understanding is correct, please add `public` accessor 🙏🏼

app-ios/Modules/Sources/Theme/Typography.swift
```diff
@@ -0,0 +1,43 @@
+import SwiftUI
+
+public struct Typography {
+    var displayLarge: TextStyle = TypographyTokens.displayLarge
```
ry-itto, line 4: please add accssor to variables as well 🙏

Corvus400:
@ry-itto 
Please review again as we have resolved all the issues raised! :bow:

ry-itto (approved):
LGTM!

PR #1136, opened 2023-09-09
[iOS] Share sessions that depend on the language setting

woxtu:
## Issue
- close N/A

## Overview (Required)
- This PR improves to share sessions that depend on the language setting

## Links
- https://github.com/DroidKaigi/conference-app-2023/blob/d88402ccb6582060a456bdf712250a66c9b4a5bd/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2023/model/TimetableItem.kt#L96-L100

Before | After
:--: | :--:
|| <br>(Japanese)
|| <br>(English)

## Movie (Optional)
Before | After
:--: | :--:


ry-itto (approved):
LGTM! Thank you!

PR #1139, opened 2023-09-09
[iOS] use StateObject instead of ObservedObject

### LINKED ISSUE #1077: [iOS] Improve ViewModel load timing for all pages
**Idea Description**

- Currently, viewModel loaded at every appeared timing. Please improve it to only at once.

ry-itto:
## Issue
- close #1077

## Overview (Required)
- Currently, iOS App loads every time if each view opened. This PR improves that behavior. (If some view opened at second time, it will not reload state.)

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


RyuNen344 (approved):
LGTM 😉

PR #1153, opened 2023-09-10
[iOS] Show empty search results properly

woxtu:
## Issue
- close N/A

## Overview (Required)
- This PR fixes a view's frame to show empty search results properly

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


ry-itto (approved):
Good catch! Thanks 🙏🏼

PR #1157, opened 2023-09-10
Replace enum values() with entries

Pinoshino2:
## Issue
- no issue
## Overview (Required)
- In Kotlin 1.9.0, the entries property of Enum is Stable. The official documentation recommends replacing values() with entries, so I have fixed as much as possible.
- it does not return an array, so a new instance is not created with each call, which may improve Kotlin performance slightly.

## Links
- https://kotlinlang.org/docs/whatsnew19.html#stable-replacement-of-the-enum-class-values-function

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):
Great!

Issue #762, opened 2023-08-19
[Error] "Type argument is not within its bounds" in Android Studio


TonTonbow:
**Describe the bug**

Errors are occurring at the following locations.
Doesn't seem to be a problem for builds, etc.

https://github.com/DroidKaigi/conference-app-2023/blob/28a7f8d63745da9201e2bbd0fe2c5b0a43ad0e52/app-android/src/main/java/io/github/droidkaigi/confsched2023/KaigiApp.kt#L188

Error Content:
```shell
Type argument is not within its bounds: should be subtype of 'ViewModel'
```

<details>
<summary>日本語</summary>

以下の箇所でエラーが発生している。
ビルドなどには影響は無いようです。
https://github.com/DroidKaigi/conference-app-2023/blob/28a7f8d63745da9201e2bbd0fe2c5b0a43ad0e52/app-android/src/main/java/io/github/droidkaigi/confsched2023/KaigiApp.kt#L188

エラー内容：
```shell
Type argument is not within its bounds: should be subtype of 'ViewModel'
```
</details>

**Expected behavior**
Error Resolution

<img width="1220" src="https://github.com/DroidKaigi/conference-app-2023/assets/70003919/c3a0f664-fa40-48da-9bef-a30ffce0a1ba">

takahirom:
It is only in Android Studio, right? I can see the same. Maybe it is KMP problem.

TonTonbow:
Yes

ked4ma:
When I open this repo with "Android Studio Giraffe | 2022.3.1 Patch 1",
It seems to be resolved. (And we might be able to remove a generics)

If you try to download AS from [here](https://developer.android.com/studio), you will get "2022.3.1 Patch 1".
And if this version AS can be considered as "STABLE", I can make a PR for it :runner: 

![Screenshot from 2023-08-28 19-40-18](https://github.com/DroidKaigi/conference-app-2023/assets/74723074/6b335c99-252c-40aa-a6e7-395f8c75cb44)
![Screenshot from 2023-08-28 19-43-28](https://github.com/DroidKaigi/conference-app-2023/assets/74723074/af869c23-d5b0-40bd-9ad1-a686ccca4028)

Nanamare:
@ked4ma Thanks your information

PR #1158, opened 2023-09-10
[Android] remove redundant type parameter

tkhs0604:
## Issue
- close https://github.com/DroidKaigi/conference-app-2023/issues/762

## Overview (Required)
- SSIA
- It has already resolved in the latest stable version.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


RyuNen344 (approved):
LGTM ❤️

PR #1161, opened 2023-09-10
:sparkles: Added a shell script to create the XCFramework needed when developing the iOS side.

Corvus400:
## Issue
- None.

## Overview (Required)
- Since I forget the command every time when outputting XCFramework, I added a script that can output the minimum required XCFramework by executing a shell script.

### DIFF
--- a/scripts/assemble_xc_framework.sh
+++ b/scripts/assemble_xc_framework.sh
@@ -1,0 +1,6 @@
+#!/bin/zsh
+
+cd ../
+./gradlew :app-io-shared:assembleSharedReleaseXCFramework
+./gradlew :app-io-shared:assembleSharedDebugXCFramework 
+echo -e "\a"


scripts/assemble_xc_framework.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/zsh
+
+cd ../
+./gradlew :app-io-shared:assembleSharedReleaseXCFramework
+./gradlew :app-io-shared:assembleSharedDebugXCFramework 
+echo -e "\a"
```
Corvus400, line 6: I specified this because I want a sound to be played when the output is completed.

scripts/assemble_xc_framework.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/zsh
+
+cd ../
+./gradlew :app-io-shared:assembleSharedReleaseXCFramework
```
Corvus400, line 4: Debug alone was not enough, so we also ran the assembleXCFramework in Release.

takahirom: Maybe the iOS app currently doesn't use a debug framework so maybe we can remove it?
It is better to be able to use it to streamline the build/

takahirom: I don't have strong opinion about this but there are fastlane script to do this. Maybe iOS enginers are familiar with that way.

takahirom: Do you have any opinion on this? @RyuNen344

Corvus400: @takahirom @RyuNen344 @ry-itto 
Our goal is to facilitate iOS development, so of course if there is a better way, we would prefer it.🙌

PR #1180, opened 2023-09-11
[iOS] fix time order and duplicate contents

### LINKED ISSUE #1165: [iOS] Fix contents and their order to align with Android
The contents and their order displayed on each tab are different from Android's ones.
They should be aligned with Android.
Bookmark and search screens could be the same situation.

&nbsp; | iOS | Android
:--: | :--: | :--:
Day1. |  | 
Day.2 |  |  
Day.3 |  |

sohichiro:
## Issue
- close #1165

## Overview (Required)
I resolved the issues with timetable time shifts and duplicate content.
It is not best solution, but now I thought that this issue have to solve immediately.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/6025572/d33709e2-f985-4ba0-b89d-e0cacefca84c" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/6025572/885b0bf7-7cbc-4f88-a99f-31d6ff34c1e6" width="300" >

### DIFF
--- a/app-ios/Modules/Sources/Timetable/TimetableViewModel.swift
+++ b/app-ios/Modules/Sources/Timetable/TimetableViewModel.swift
@@ -68,14 +68,16 @@ final class TimetableViewModel: ObservableObject {
                         $0.timetableItem.room.name.currentLangTitle < $1.timetableItem.room.name.currentLangTitle
                     }
                 return TimetableTimeGroupItems(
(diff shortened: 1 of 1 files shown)


ry-itto (approved):
LGTM! Thank you!

PR #1182, opened 2023-09-11
[iOS] Fix the tab title for Floor Map

woxtu:
## Issue
- close N/A

## Overview (Required)
- This PR makes space to fix the tab title for Floor Map

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


ry-itto (approved):
LGTM!

PR #1193, opened 2023-09-12
[iOS] Localize bookmarks

woxtu:
## Issue
- close N/A

## Overview (Required)
- This PR localizes Bookmarks

## Links
- 

Before | After
:--: | :--:
|| <br>(Japanese)
|| <br>(English)

## Movie (Optional)
Before | After
:--: | :--:


app-ios/Modules/Sources/Timetable/Bookmark/BookmarkView.swift
```diff
@@ -48,6 +48,7 @@ struct BookmarkView<SessionView: View>: View {
                             )
                         }
                     }
+                    .padding(.horizontal, 16)
```
tkhs0604, line 51: 👍

tkhs0604 (approved):
Thank you as always! LGTM 🙌

PR #1196, opened 2023-09-12
:memo: As last year, we replaced the URL with the store's URL.

Corvus400:
## Issue
- None.

## Overview (Required)
- Since the app is already available in the store, we have replaced it with a link to the store as we did last year.

## Links
- https://github.com/DroidKaigi/conference-app-2022#android


Corvus400:
Oops...it was already addressed, so I'll close it. 🙇

PR #1197, opened 2023-09-12
[iOS] :wrench: Nothing is displayed when side event images cannot be loaded.

Corvus400:
## Issue
- None.

## Overview (Required)
- Figma also does not seem to display anything when an image does not exist, so we changed the appearance of the image when it could not be loaded accordingly.

## Links
- https://www.figma.com/file/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?type=design&node-id=56145-69943&mode=dev

Before | After
:--: | :--:
 |


ry-itto (approved):
LGTM! Thanks!

PR #1198, opened 2023-09-12
[iOS] delete duplicate session

sohichiro:
## Issue
- nothing

## Overview (Required)
- In my environment, there was duplicated content on the search screen, so I have implemented this so that there is no duplicated content.

## Links
- 
## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/6025572/46740aca-4a01-4a67-9550-f91968ef17ed" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/6025572/87f98a02-9873-4016-80ae-8e57d2f9ef4d" width="300" >


ry-itto (approved):
LGTM!

PR #1202, opened 2023-09-12
[iOS] :wrench: Remove duplicate elements from the language selection on the search screen.

Corvus400:
## Issue
- None.(Maybe)

## Overview (Required)
- When trying to filter supported languages on the search screen, duplicate languages were being displayed, so we have made them non-duplicate.
- The duplicate languages were those with isInterpretationTarget true and false, respectively.
- I am not sure if this response is correct.
- Would it be better to modify it to clearly state that it is a Japanese/English session that supports simultaneous translation?

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/da2ae5e5-8be5-438d-b80e-6e70a08bfac2" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/13657682/976c8edf-5eab-4913-93b9-39db7ed060e7" width="300" >


ry-itto (approved):
Good catch! Thanks 🙏🏼

PR #1213, opened 2023-09-14
give suitable margins for the timetable grid

urotarou:
## Issue
- none

## Overview (Required)
- The padding given by Scaffold was not applied to the bottom of the timetable grid, and the bottom of the last session was missing.
- Timetable We thought that applying padding to the end of the grid would result in extra padding, so we thought it would be appropriate to give padding only to the bottom and start.

## Links
- none

Before | After
:--: | :--:
 | 

## Movie (Optional)
- none


takahirom (approved):
I think this is what we need! Thanks

PR #1214, opened 2023-09-14
changed bottom navigation return behavior

urotarou:
## Issue
- none

## Overview (Required)
- Pop up to the start destination of the graph to avoid building up a large stack of destinations on the back stack as users select items.
- Sorry if the current behavior is the expected specification.

## Links
- https://developer.android.com/jetpack/compose/navigation?hl=ja#bottom-nav

- none

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/8335257/2126ded0-c71f-464d-b532-8364a4752741" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/8335257/76fb25d4-4c67-4a7b-8ab6-3fe320d44ce6" width="300" >


takahirom (approved):
Looks great!

PR #1215, opened 2023-09-14
Use compose bom version 2023.09.00

matsumo0922:
## Issue
- N/A 

## Overview (Required)
- Changed version control of Compose to be done by Bom
- Compose Bom version is set to 2023.09.00

## Links
- https://developer.android.com/jetpack/compose/bom


takahirom:
Maybe the font is changed. Do you know which font is the right font?

takahirom:
Well, maybe the font is the same but the space of the text is changed?

takahirom (approved):
Thanks for aligning dependencies! This is what we need!

PR #1216, opened 2023-09-14
Fix timetable detail section loading animation

matsumo0922:
## Issue
- TimetableDetailScreen animation is unnatural.

## Overview (Required)
- The contentKey was introduced in #982, but during the initial animation, there is a moment when the content is not displayed, and the loaded section is drawn from there, resulting in a scaled animation that is unnatural.
- Therefore, only the Loading section is faded animated without AnimatedContent.
  - In the first place, this loading never takes more than a few seconds (for local), so I feel that the switching animation is not necessary.

## Movie
Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/56629437/1819b828-9f4e-4727-8dda-2c7bd34b8d71" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/56629437/df006910-86c3-4487-99e5-7ba553f57bea" width="300" >


takahirom (approved):
I've been concerned about this! Thanks!

Issue #1219, opened 2023-09-15
Adjusting Time Settings to "Day 1" for Screenshot Tests


takahirom:
**Idea Description**
In the current implementation of our screenshot tests, the time settings are automatically set to "Day 2." This has resulted in some inconsistencies in the test outcomes. The purpose of this document is to outline the steps required to adjust the time settings to "Day 1" for more accurate and consistent screenshot tests.

**Reference images and links**

kubode:
🤚 
I'll try this.

takahirom:
@kubode Thanks. But there is a pr that trying this. It seems that it doesn't work though 👀 
https://github.com/DroidKaigi/conference-app-2023/pull/1224

kubode:
Thanks! That PR looks good 👏

PR #1224, opened 2023-09-15
[Android] add LocalClock to control time for screenshot test

tkhs0604:
## Issue
- close https://github.com/DroidKaigi/conference-app-2023/issues/1219

## Overview (Required)
- SSIA
- Adding a composition local for clock makes it slightly more adjustable to set datetime information for screenshot test

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


core/ui/src/androidMain/kotlin/io/github/droidkaigi/confsched2023/ui/compositionlocal/LocalClockProvider.kt
```diff
@@ -0,0 +1,9 @@
+package io.github.droidkaigi.confsched2023.ui.compositionlocal
+
+import androidx.compose.runtime.compositionLocalOf
+import kotlinx.datetime.Clock
+
+@Suppress("CompositionLocalAllowList")
+val LocalClockProvider = compositionLocalOf<Clock> {
```
kubode, line 7: IMO, using `staticCompositionLocalOf` would be better performance.

kubode (commented):
I tried to work on https://github.com/DroidKaigi/conference-app-2023/issues/1219 but was told that the PR is already there.
I commented where I was concerned.

core/ui/src/androidMain/kotlin/io/github/droidkaigi/confsched2023/ui/compositionlocal/LocalClockProvider.kt
tkhs0604: Thank you for the advice! Let me check the function and replace with it!

tkhs0604:
Hmm... The format error cannot be resolved even if I suppressed it 🫠...

tkhs0604:
x: @Suppress("CompositionLocalAllow**L**ist")
o: @Suppress("CompositionLocalAllow**l**ist")
😢

core/testing/src/main/java/io/github/droidkaigi/confsched2023/testing/robot/TimetableScreenRobot.kt
```diff
@@ -50,12 +53,14 @@ class TimetableScreenRobot @Inject constructor(
 
     fun setupTimetableScreenContent() {
         composeTestRule.setContent {
-            KaigiTheme {
-                TimetableScreen(
-                    onSearchClick = { },
-                    onTimetableItemClick = { },
-                    onBookmarkIconClick = { },
-                )
+            CompositionLocalProvider(LocalClock provides FakeClock) {
```
tkhs0604, line 56: It does not seem like it works well...
How can I overwrite the CompositionLocal in test environment? 🤔

Maybe, since there is no setContent in KaigiAppTest, so the only way is to add like ClockTestRule or something like that?

tkhs0604, line 56: @takahirom 
It does not seem like it works well...
How can I overwrite the CompositionLocal in test environment? 🤔

Maybe, since there is no setContent in KaigiAppTest, so the only way is to add like ClockTestRule or something like that?

takahirom: I'm not sure too. 
By the way, you might have to fix the commit history 👀

tkhs0604: Thank you, I'll investigate the cause again 🙏
And regarding your BTW comment, which commits? All?

takahirom: Sorry, I might see old commit 🙇

tkhs0604: The FakeClock worked well for only TimetableScreenTest on my local, but the screenshot results haven't been updated even if I added some commits 👀
Is there any way to update them on this PR's comment?

In case of `2023-09-14T10:00:00.000Z`
TimetableScreenTest checkGridScrollShot_compare

In case of `2023-09-15T10:00:00.000Z` (To check the diff)
TimetableScreenTest checkGridScrollShot_compare

takahirom: I've removed the comment. Let's see what screenshots the tests take.

takahirom: image
It seems that there is no change 🎉

takahirom (approved):
Thanks!

core/testing/src/main/java/io/github/droidkaigi/confsched2023/testing/robot/TimetableScreenRobot.kt
tkhs0604: Thank you! 🙌

PR #1240, opened 2023-09-16
Fix the issue of TimetableGrid is off by one hour

### LINKED ISSUE #1226: Time Display on TimetableGrid is Off by One Hour
**Idea Description**
The time displayed on the TimetableGrid is one hour behind the actual scheduled time.

image

For now, we released a hotfix veresion that we remove the time as 1.6.1.

image

aqua-ix:
## Issue
- close #1226 

## Overview (Required)
- In order to fix the issue where TimetableGlid was off by one hour, fixed the base time for TimeTableGlid calculation to 10:00 JST.

Before | After
:--: | :--:
 |


takahirom (approved):
Ideally, it would be better to have the start time set dynamically. But it looks great!

PR #1245, opened 2023-09-16
Migrate dagger hilt from kapt to ksp.

### LINKED ISSUE #1234: KSP Dagger challenge
**Idea Description**
Dagger, which is DI library has been updated to support KSP. Could we use the support?

CookieySun:
## Issue
- close #1234

## Overview (Required)
- Migrate dagger hilt from kapt to ksp.


takahirom (approved):
Great!

PR #1265, opened 2023-09-17
Add achivement click animation #1235

### LINKED ISSUE #1235: Allow Replay of Achievement Animations Only for Achievements Already Acquired
**Idea Description**
The feature request is to enable users to replay the animation of achievements they've already acquired. Users should only be able to replay animations for achievements listed in their "Achievements" section as already obtained.

Aniokrait:
## Issue
- close #1235 

## Overview (Required)
- add animation on click for acquired achivements.
- refactor and make a component for achivement animation.
- It looked more natural with the background slightly transparent, so I made the background translucent.

## Movie (Optional)
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/43056135/09c5c592-a851-49be-b7d1-6b96769bc3dc" width="300" >

### DIFF
--- a/feature/achievements/src/main/java/io/github/droidkaigi/confsched2023/achievements/AchievementsScreen.kt
+++ b/feature/achievements/src/main/java/io/github/droidkaigi/confsched2023/achievements/AchievementsScreen.kt
@@ -21,6 +21,7 @@ import androidx.compose.material3.SnackbarHostState
 import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
@@ -150,6 +151,11 @@ private fun AchievementsScreen(
             },
         )
         if (uiState.clickedAchievement is Clicked) {
+            DisposableEffect(uiState.clickedAchievement) {
+                onDispose {
+                    finishAnimation()
+                }
+            }
             Surface(
                 modifier = Modifier.fillMaxSize(),
                 color = MaterialTheme.colorScheme.background.copy(alpha = 0.6F),
--- a/feature/achievements/src/main/java/io/github/droidkaigi/confsched2023/achievements/AchievementsScreen.kt
+++ b/feature/achievements/src/main/java/io/github/droidkaigi/confsched2023/achievements/AchievementsScreen.kt
@@ -1,12 +1,14 @@
 package io.github.droidkaigi.confsched2023.achievements
 
 import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.calculateEndPadding
 import androidx.compose.foundation.layout.calculateStartPadding
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.rememberScrollState
@@ -16,6 +18,7 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Scaffold
 import androidx.compose.material3.SnackbarHost
 import androidx.compose.material3.SnackbarHostState
+import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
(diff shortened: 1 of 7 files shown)


feature/achievements/src/main/java/io/github/droidkaigi/confsched2023/achievements/AchievementsScreen.kt
```diff
@@ -83,51 +90,85 @@ fun AchievementsScreen(
...
     )
 }
 
 data class AchievementsScreenUiState(
     val achievementListUiState: AchievementListUiState,
     val isShowInitialDialog: Boolean,
+    val clickedAchievement: ClickedAchievementState,
 )
 
+sealed interface ClickedAchievementState {
```
takahirom, line 105: How about naming this AchievementAnimationState and AchievementAnimationState.Animating AchievementAnimationState.NotAnimating?

Aniokrait: The namings look better. I will rename them.

feature/achievements/src/main/java/io/github/droidkaigi/confsched2023/achievements/AchievementsScreen.kt
```diff
@@ -83,51 +90,85 @@ fun AchievementsScreen(
         snackbarHostState = snackbarHostState,
         contentPadding = contentPadding,
         onReset = viewModel::onReset,
+        showAnimation = { achievement -> viewModel.onClickAchievement(achievement) },
+        finishAnimation = viewModel::onFinishAnimation,
```
takahirom, line 94: We would like to promote the best practice that a StateHolder should be notified events. 
So how about naming this `onAchivementClicked` and judge if we run a animation in ViewModel for `showAnimation`?
And use "onAnimationFinished" as a name for `finishAnimation`?

feature/achievements/src/main/java/io/github/droidkaigi/confsched2023/achievements/AchievementsScreenViewModel.kt
```diff
@@ -120,13 +121,17 @@ class AchievementsScreenViewModel @Inject constructor(
         )
     }
 
+    private val clickedAchievement = MutableStateFlow<AchievementAnimationState>(AchievementAnimationState.NotAnimating)
```
takahirom, line 124: Could you change this name as we change the state? 🙇

takahirom (commented):
I would like to have a screenshot for this ideally. Could you see it? Maybe we can add a archivement by injecting the datastore in the test. 👀

feature/achievements/src/main/java/io/github/droidkaigi/confsched2023/achievements/AchievementsScreen.kt
Aniokrait: Renamed showAnimation to onAchivementClick and finishAnimation to onAnimationFinish.

feature/achievements/src/main/java/io/github/droidkaigi/confsched2023/achievements/AchievementsScreenViewModel.kt
Aniokrait: Sorry for overlooking. I renamed clickedAchievement to achievementAnimationState.

Aniokrait:
> I would like to have a screenshot for this ideally. Could you see it? Maybe we can add a archivement by injecting the datastore in the test. 👀

I've tried to add the screenshot tests. It seems to work fine.

feature/achievements/src/test/java/io/github/droidkaigi/confsched2023/achievements/AchievementsScreenTest.kt
```diff
@@ -30,6 +38,11 @@ class AchievementsScreenTest {
     @Inject
     lateinit var achievementsScreenRobot: AchievementsScreenRobot
 
+    @Before
+    fun setup() {
+        LottieTask.EXECUTOR = Executor(Runnable::run)
```
takahirom, line 43: Maybe you should restore the executor after test to isolate the tests 👀

Aniokrait: @takahirom 
I've checked if this setup affects other test classes.
To invesigate that, I created AchievementsScreenTest2 copying original one except setUp function. As a result, all the test cases in AchievementsScreenTest2 didn't capture achivement highlight correctly because LottieTask.EXECUTOR was not set up for tests.
So, I assume changing LottieTask.EXECUTOR is confined in its test class and tearDown isn't really necessary.
However, if you are concerned about other test cases in the same test class, which doesn't need to set up LottieTask.EXECUTOR, I'd make setup in each test cases like below. But I feel it's kind of hacky.
```
fun checkHighlightImageA() {
    achievementsScreenRobot {
        LottieTask.EXECUTOR = Executor(Runnable::run)
        setupSavedAchievement(ArcticFox)
        setupScreenContent()
        clickAchievementImageA()
        checkScreenCapture()
        LottieTask.EXECUTOR = Executors.newCachedThreadPool()
    }
}
```

takahirom: As far as I know static variable is used across the tests. We have some options to do it. 
How about adding it here?
https://github.com/DroidKaigi/conference-app-2023/blob/b53057115722cacf1a7fcd6eee2786d1a0267516/core/testing/src/main/java/io/github/droidkaigi/confsched2023/testing/RobotTestRule.kt#L64

Aniokrait: @takahirom 
I wasn't aware of that system.
I added LottieTestRule.

PR #1273, opened 2023-09-23
Support fling for nested scroll on TimetableGrid.

### LINKED ISSUE #776: Nested Scroll on TimetableGrid should support fling operation.
**Idea Description**

In the Timetable grid section of the DroidKaigi app, the NestedScroll functionality is implemented.
When scrolling TimetableGrid, the header and the tab shrinks before / expands after TimetableGrid scrolls itself.
But so far fling process is not implemented.
For more smooth scroll experience, nested fling should be implemented.

**Expected Behavior**

After up scrolling TimetableGrid and releasing the finger, the header and the tab should shrink before Timetable fling process.
After down srcolling, the header and the tab should expand after Timetable scrolls to the edge.

**Current Behavior**

After scrolling TimetableGrid and releasing the finger, TimetableGrid will continue to scroll for a while, but the header and the tab will not shrink or expand any further.

- #498 
- #659

usuiat:
## Issue
- close #776 

## Overview (Required)
- Support fling operation for nested scroll on TimetableGrid.

After drag gesture on TimetableGrid ends, the header height and the tab height expand or shrink in conjunction with TimetableGrid flinging.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/17829294/a1c27972-6b22-48ef-9cc1-d51f363830b0" width="300" > | <video src="https://github.com/DroidKaigi/conference-app-2023/assets/17829294/5952d37e-8f57-4148-b60d-5977322ab152" width="300" >

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/section/TimetableGrid.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/section/TimetableGrid.kt
@@ -35,8 +35,10 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clipToBounds
 import androidx.compose.ui.draw.drawBehind
 import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.input.nestedscroll.NestedScrollConnection
 import androidx.compose.ui.input.nestedscroll.NestedScrollDispatcher
 import androidx.compose.ui.input.nestedscroll.NestedScrollSource
+import androidx.compose.ui.input.nestedscroll.nestedScroll
 import androidx.compose.ui.input.pointer.PointerInputChange
 import androidx.compose.ui.input.pointer.PointerInputScope
 import androidx.compose.ui.input.pointer.pointerInput
@@ -84,14 +86,12 @@ data class TimetableGridUiState(val timetable: Timetable)
 @Composable
 fun TimetableGrid(
     uiState: TimetableGridUiState,
-    nestedScrollDispatcher: NestedScrollDispatcher,
     onTimetableItemClick: (TimetableItem) -> Unit,
     modifier: Modifier = Modifier,
     contentPadding: PaddingValues = PaddingValues(),
 ) {
     TimetableGrid(
         timetable = uiState.timetable,
-        nestedScrollDispatcher = nestedScrollDispatcher,
         onTimetableItemClick = onTimetableItemClick,
         modifier = modifier,
         contentPadding = contentPadding,
@@ -101,7 +101,6 @@ fun TimetableGrid(
 @Composable
 fun TimetableGrid(
     timetable: Timetable,
-    nestedScrollDispatcher: NestedScrollDispatcher,
     onTimetableItemClick: (TimetableItem) -> Unit,
     modifier: Modifier = Modifier,
     contentPadding: PaddingValues = PaddingValues(),
@@ -133,7 +132,6 @@ fun TimetableGrid(
             TimetableGrid(
                 timetable = timetable,
                 timetableState = timetableGridState,
-                nestedScrollDispatcher = nestedScrollDispatcher,
                 modifier = modifier,
                 contentPadding = PaddingValues(
                     top = 16.dp + contentPadding.calculateTopPadding(),
@@ -157,7 +155,6 @@ fun TimetableGrid(
 fun TimetableGrid(
     timetable: Timetable,
     timetableState: TimetableState,
-    nestedScrollDispatcher: NestedScrollDispatcher,
     modifier: Modifier = Modifier,
     contentPadding: PaddingValues = PaddingValues(),
     content: @Composable (TimetableItem, Int) -> Unit,
@@ -186,10 +183,14 @@ fun TimetableGrid(
         content(timetableItemWithFavorite.timetableItem, itemHeightPx)
     }
 
+    val nestedScrollConnection = remember { object : NestedScrollConnection {} }
+    val nestedScrollDispatcher = remember { NestedScrollDispatcher() }
+
     LazyLayout(
         modifier = modifier
             .focusGroup()
             .clipToBounds()
+            .nestedScroll(nestedScrollConnection, nestedScrollDispatcher)
             .drawBehind {
                 timetableScreen.timeHorizontalLines.value.forEach {
                     drawLine(
@@ -322,7 +323,6 @@ fun TimetableGrid(
 fun TimetablePreview() {
     TimetableGrid(
         timetable = Timetable.fake(),
-        nestedScrollDispatcher = remember { NestedScrollDispatcher() },
         onTimetableItemClick = {},
         modifier = Modifier.fillMaxSize(),
     )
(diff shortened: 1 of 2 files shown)


feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/section/TimetableSheet.kt
```diff
@@ -128,17 +127,11 @@ fun TimetableSheet(
                 }
 
                 is GridTimetable -> {
-                    val nestedScrollDispatcher = remember { NestedScrollDispatcher() }
```
swimmy-reo, line 131: Thanks for the great PR!
Is there any reason why you originally applied the NestedScrollDispatcher definition and Modifier nestedScroll here?

usuiat: At first I was misunderstanding that I needed to pass a nestedScrollDispatcher and a timetableSheetContentScrollState.nestedScrollConnection to one Modifier.nestedScroll() argument. 
But in this case, the TimetableGrid's nestedScrollDispatcher is independent of the TimetableSheet. Therefore I moved the dispatcher into TimetableGrid class.

swimmy-reo (commented):
I was a little curious about the behavior when I moved it in my hand, so let me share it with you.
（We are still in the process of figuring out which part of the problem can be resolved, so we will only share once.）

First, is it an expected behavior that the scroll position returns to the original position after a slight fling at the initial position?

Second, when I scroll in the direction that the border wraps around, it scrolls in the opposite direction. Is this unintended behavior?

(This may just be on my end, so please let me know if this does not reproduce :pray:)

1st | 2nd
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/65647834/07c84eb9-d6c1-4792-b74a-e45c183639e7" width="300" />|<video src="https://github.com/DroidKaigi/conference-app-2023/assets/65647834/68f0cf08-b81c-4cce-9eea-d7a8077e360e" width="300" >

usuiat:
@swimmy-reo 
> First, is it an expected behavior that the scroll position returns to the original position after a slight fling at the initial position?
Second, when I scroll in the direction that the border wraps around, it scrolls in the opposite direction. Is this unintended behavior?

Thank you for pointing out.
These are not expected behaviors but I can reproduce them.

I referred to the Modifier.scrollable() code when implementing this fling operation. 
So I will check the code carefully again.

usuiat:
@swimmy-reo 
I have fixed some not good code.
Could you check again? Thank you.

feature/sessions/src/main/java/io/github/droidkaigi/confsched2023/sessions/section/TimetableGrid.kt
```diff
@@ -672,6 +713,9 @@ private class TimetableScreen(
         position: Offset,
         nestedScrollDispatcher: NestedScrollDispatcher,
     ) {
+        // If the position does not change, VelocityTracker malfunctions. Therefore return here.
+        if (dragAmount == Offset.Zero) return
```
swimmy-reo, line 717: Wow! I did not know that! 😮

swimmy-reo:
🏅 🏅 🏅 🏅 I was amazed at how quickly the correction was handled! Amazing! 🏅 🏅 🏅 🏅 

Let me make one suggestion.
I was concerned about the fact that scrolling in the grid is no longer working. I think it would be better to be able to scroll as there is a certain number of contents.

I think it would be good to see if the tabs are collapsed when switching the cancelFling flag that you defined.

current  | ideal
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/65647834/1cbb8b32-6afc-4134-bcc7-2a5aab8faa87" width="300" />|<video src="https://github.com/DroidKaigi/conference-app-2023/assets/65647834/07db2d78-0062-4cfd-8922-80e2cb3172aa" width="300" >

```
// In TimetableGrid.kt

// add isTabExpandable property
data class TimetableGridUiState(val timetable: Timetable, val isTabExpandable: Boolean? = null )

@Composable
fun TimetableGrid(
    timetable: Timetable,
    isTabExpandable: Boolean?, // new
    onTimetableItemClick: (TimetableItem) -> Unit,
    contentPadding: PaddingValues = PaddingValues(),
) {
      ...
 }


@OptIn(ExperimentalFoundationApi::class)
@Composable
fun TimetableGrid(
    timetable: Timetable,
    isTabExpandable: Boolean?, // new
    timetableState: TimetableState,
    contentPadding: PaddingValues = PaddingValues(),
    content: @Composable (TimetableItem, Int) -> Unit,
) {
      ...
}

@Stable
class ScreenScrollState(
    initialScrollX: Float = 0f,
    initialScrollY: Float = 0f,
) {
    ...
    suspend fun scroll(
        scrollX: Float,
        scrollY: Float,
        timeMillis: Long,
        position: Offset,
        isTabExpandable: Boolean? = null, // new
    ) {
        if (isTabExpandable == false) cancelFling = true // fixed
       ...
      }
   }
```

To use CompositionLocal, I'm thinking of passing the data in a bucket relay format since the data is of limited use.

As before, let me know if you can't reproduce this!
Also, please let me know if I said something wrong!


It's very difficult to verbalize what the ideal behavior of nested scrolling would be, so it's tough to know without trying to modify it. 😅

usuiat:
@swimmy-reo 
I have tried but cannot reproduce it.
Is fling not working at all? On my end, it is generally working.

Yes, sometimes the fling does not work (scrolling stops as soon as I release my finger), but I think this is the same as before. (This is not a good, though 🤔)

usuiat:
@swimmy-reo 
If you are running the app in debug variant, could you try release (devRelease) variant?
In my environment, the scrolling tends to stop with debug variant. However, in the release variant the scrolling runs relatively smoothly.
If you are the same, this may be a performance issue.

swimmy-reo:
I see!
I see that the fling is working properly in usui's hand...
I'll check it out.

swimmy-reo:
I'm having someone else check it out to see if it works.
If it works correctly, I would like to approve it as is!

To be honest, my device specs are low lol.
If there are more devices to reproduce, I would like to consider the performance aspect.

swimmy-reo:
I will reply this evening!

swimmy-reo:
I have had others look at this and the degree of reproduction seems to vary from device to device.

Below is a release build with Pixel Fold.
When scrolling from bottom to top, the fling does not work, and when scrolling from top to bottom, the fling works.
pixel fold  |
:--: | 
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/65647834/80bd8d84-51e1-44d5-b9fe-5b71c8485a72" width="300" />|

We asked Pixel Fold to debug in detail the point where the fling does not work when scrolling from bottom to top, and found that the fling does not work when scrolling in the direction where there is no grid, but it works when scrolling in the direction where there is a grid.

no fling  | fling
:--: | :--:
<video src="https://github.com/DroidKaigi/conference-app-2023/assets/65647834/cfb4f09a-ae6a-4202-968a-e974673a17aa" width="300" />|<video src="https://github.com/DroidKaigi/conference-app-2023/assets/65647834/212ee697-471b-41ee-b82c-1e7056492448" width="300" />

Here we see that the reproducibility varies from device to device.
Since addressing this device difference is outside the scope of this PR, we would like to isolate it as a separate issue and approve it here!

[Here is the issue.](https://github.com/DroidKaigi/conference-app-2023/issues/1276)

usuiat:
I appreciate your sincere review and people who helped👏 
I learned a lot working on this Issue.

RyuNen344 (approved):
❤️

RyuNen344:
@usuiat 
Could you merge this PR if you can ?

usuiat:
@RyuNen344 
I don't think I have the authority to merge.

